PR #60, opened 2014-11-04
Add es6-promise to registry

computmaxer:
Also add package-override for version 2.0.0.

@guybedford


trusktr:
@guybedford Interestingly I'm not the only one who thought directories.lib could be used on an npm package. Could `files` be used for my `src` problem?

guybedford:
Closing for now, an updated PR is welcome.

PR #96, opened 2015-01-03
added es5-shim and es6-shim

Aaike:
Added es5 and es6 shims, no override files


guybedford:
Thanks, I've included these in https://github.com/jspm/registry/commit/07fbb703a73da7463a03cdb520d59850d9c9ebbe.

PR #98, opened 2015-01-05
Update famous. Adde override to my form of famous.

trusktr:


guybedford:
There is no reason to have `trusktr/famous` here as an override - rather include that configuration in the `trusktr/famous` package.json directly.

trusktr:
@guybedford The reason I haven't done that is because I'm using this same branch to make PRs to famous, and they didn't want the jspm section in package.json. I can make a new branch just for it though. Let me do that.

trusktr:
Yeah, let me update famous to the npm version too. Seems like a better choice.

trusktr:
Aha!!! I remember why I chose the github version of famous!! for some reason, jspm is installing some version famous@11.18.0, which completely doesn't exist as far as http://famo.us is concerned. So, when my project uses that lib, of course it doesn't work.

`jspm install famous=npm:famous` is putting `famous@^11.18.0` in package.json, but I can manually change it to `famous@0.3.4` and it will work.

trusktr:
I suppose I'll also leave `infamous` as a github package since with the npm version I can't bring stuff from src into the root using directories.lib?

trusktr:
@guybedford There we go. Removed my famous fork override. I'm still using github for both famous and my fork because then the directories.lib thing works with both, and also due to the famous version problem.

guybedford:
I've created an issue for the point you mentioned at https://github.com/jspm/npm/issues/34, thanks.

What is the repo at `trusktr/famous@jspm` for? Is it for public consumption? It is best not to have private repos in the registry so let me know if we should remove this at some point in future.

trusktr:
@guybedford The branch at `trusktr/famous@jspm` is just my customized version of famous with package.json modified for jspm consumption. I suppose this could go in a custom registry, but I haven't tried that yet. Being able to save overrides in package.json would make this easier. I'm just storing the override for `trusktr/famous@jspm` in this registry because it was easy, but editing a project's package.json would be the easiest, and not add extra stuff to this registry.

trusktr:
I guess I don't really need it any more. I can just do `jspm install -y famous=github:trusktr/famous@jspm`.

famous unpublish the extraneous `famous@11.18.0` so now `jspm install famous` install the proper version, but this leaves all the files in the `src` folder. How can we get around this?

trusktr:
Ah, PR https://github.com/jspm/jspm-cli/pull/372 is how. xD

PR #99, opened 2015-01-05
Added infamous name.

trusktr:


trusktr:
I'd use the npm version here too, but then I won't have the benefits of directories.lib.

guybedford:
I thought one of the goals of this library was to provide the same API across npm etc? npm doesn't have `directories.lib`, because you just create folders for features instead:

```
main.js
featurea/index.js
featureb/index.js
```

require('pkg/featurea')

etc.

trusktr:
Well this library is only for browsers, and the intended usage in all cases is to require/import classes/modules of using the format `infamous/<path/to/module>` instead of `infamous/src/<path/to/module>`. The `src` folder just exists for organization while developing.

So for browserify and webpack, I simply compile infamous from ES6 to CJS format into the root folder on postinstall, so that works out.

For bower + requirejs, people can easily set a path in the requirejs config. Maybe I can come up with something so that's not needed.

With jspm, it's easy to do this using the github repo, and directories.lib.

As for nodejs usage, this library isn't intended for that, so I haven't thought about it.

For `npm:infamous` usage, PR https://github.com/jspm/jspm-cli/pull/372 would make this super easy to solve the problem, by letting me specify `"postinstall": "mv src/* ./"` in jspm.scripts.

guybedford:
@trusktr if you read my suggested pattern that does away with the src folder by using separate folders as modules at the base level. Let me know if that could work for you?

trusktr:
Well I just want my code away from all the book keeping files that are in the root like .gitignore, .npmignore, package.json, bower.json, etc, etc. I also don't want to have `index.js` file names. I'd like to have files names as the classes they export. I think it's nice that way. So in the `src` folder all there is to be seen is organized code.

guybedford:
Would you consider then having a `package.json` in your src folder and publishing that folder to npm perhaps then?

trusktr:
I'd like not to. Things should be flexible enough to let the dev figure out what structure is best.

guybedford:
The point here is that jspm provides the same interface as npm for npm repos. There is no reason to change it in jspm if this is something that is an npm issue. I'd rather solve this in npm first.

trusktr:
That seems like it will slow down progress. Is it really necessary? What would be a use case where not following the exact npm interface would be bad, as relating to the functionality in question here?

If jspm should provide the same npm interface, then shouldn't jspm run postinstall scripts? Arguably, the scripts needed for jspm would be different than the ones needed for npm, so that would be a reason for jspm to only run scripts found explicitly in the `jspm` section of package.json.

guybedford:
If an npm package depends on another npm package, it also expects that other npm package to have a certain folder structure. We can't just change that folder structure externally as that is breaking the contract npm packages have on eachother.

Yes the jspm npm endpoint should run npm postinstall scripts, but that doesn't mean I think it's a good idea or that the rest of jspm should.

trusktr:
Yeah, the postinstall script would be in infamous for itself, not in the parent. Though in npm it's completely possible for a parent to modify it's dependencies, but it works out since dependencies are stored per repo (so there's duplicates). That's a bit different than having all deps at the same level, no duplicates, like jspm.

trusktr:
Yeah, that could get messy. I totally am not doing that. I'm intending for infamous to modify itself, not other deps.

tauren:
I recently tried to use a project that depended on famous but wasn't
working properly. This was my first exposure to famous. While trying to
debug things, I realized that famous was doing something funky with the
directory structure and require paths. There was a src directory in the
path but the require statements didn't include src. To me, it seemed like a
wierd thing to do, a big code smell, and I quickly gave up debugging that
project and moved on to look for something better. I typically would have
fixed the problem and submitted a PR, but in this case I quickly lost
interest and didn't bother.

My point is that I personally believe all jspm packages should work exactly
as npm modules are expected to work. If they don't, you'll have confused
users and they will dump it and move on.

trusktr:
@tauren Well, even with npm, famo.us still has the `src` folder in the way, so that wouldn't work either. There are three expected ways to use famo.us (that I know of):

1. **With RequireJS:** make a [paths configuration]'(http://requirejs.org/docs/api.html#config-paths) to map `famous/src/*` to `famous/*`.
1. **With Browserify:** add the `famousify` package to your browserify project and it will automagically take care of it.
1. **as a global:** Change things like `var Engine = require('famous/core/Engine')` to `var Engine = famous.core.Engine`.

It could be arguable that famo.us could use a `postinstall` script to move all the stuff from `src` into the root after `npm install` so that neither of the requirejs and browserify steps are needed.

^ @thealphanerd

trusktr:
@tauren With a little more package management experience, these steps will be a no-brainer. ;)

trusktr:
@tauren has a good point: Some people will lose interest on things if it takes too long to set it up and if alternatives exist. I mention it because jspm is powerful and unique and does what no other package manager can do, but it currently has a higher learning curve.

PR #100, opened 2015-01-07
JSFromHell, bignumber.js, and big.js

trusktr:


registry.json
```diff
@@ -48,6 +48,8 @@
   "backbone": "npm:backbone",
   "backbone.babysitter": "github:marionettejs/backbone.babysitter",
   "backbone.wreqr": "github:marionettejs/backbone.wreqr",
+  "big.js": "github:MikeMcl/big.js",
+  "bignumber.js": "github:MikeMcl/big.js",
```
guybedford: It is best to only call something by one name in the registry. Can we go with one of `big.js` or `bignumber.js`. Also do we really need the `.js` in the name here?

trusktr: Oops, that's a typo, they are supposed to be two different repos. The repo
names have .js so I just left it. I'll be back tomorrow to fix it.

/#!/JoePea

trusktr:
@guybedford There we go, fixed it. What do you think of the aliases?

guybedford:
Let's just stick with a single name for each registry entry.

guybedford:
I'd rather focus on discovery than having lots of aliases.

Perhaps something like - 

```
  jspm registry-ls big
  big
  bignumber
```

trusktr:
Fixed. I left the `.js` since that's what the author used for the names. Someone else could claim just big or bignumber. Does that work?

guybedford:
Ok, that makes sense.

Also should we not be using the npm versions of these repos?

trusktr:
We could. What would be the benefit?

guybedford:
It's always better to use npm repos, because if other npm packages depend on them too, we have deduping in a large tree.

trusktr:
I noticed a ton of packages get installed with npm repos. I guess like build stuff, and all kinds of things I'm never going to use.

guybedford:
What gets installed is the node core libs used in a package. The overhead is being reduced on these, and shouldn't be used as an argument against.

Sometimes the node libs get detected from example folders that might use `fs`, when the main package doesn't need fs. This can be avoided by adding an `ignore: ['example']` where the example folder is `./example/`. So there are ways to reduce the overhead.

trusktr:
Changed to npm. I noticed deduping sometimes chooses lower version of deps.

guybedford:
@trusktr deduping only chooses lower version of deps when it can't be updated otherwise in the tree due to a lock within another dependency. Preference is latest, best replacement, then fork.

trusktr:
Oh, so e.g. if one dep depends on `^0.1.x` of some other dep,  and another dep depends on `^0.2.x` of the same dep, then what happens? Does that get forked? ANd does forking mean that each library loads it's version, so you get two version of the same library running in the environment?

guybedford:
Yes exactly.

PR #104, opened 2015-01-11
Add twitter

theefer:


theefer:
Actually, loading this library seems to fail with:

```
err  Error: Error loading "keygrip" at file:/home/seb/.../keygrip.js
     Error loading "keygrip" from "2-async/async" at file:/home/seb/.../2-async/async.js
     ENOENT, open '/home/seb/.../keygrip.js'
```

theefer:
Alright the `"keygrip"` library is triggering a bug in SystemJS, this PR can be merged independently but I'll open a bug against SystemJS.

PR #109, opened 2015-01-14
i18next registry and @1.7.7 override

jerrytk:
http://i18next.com/

### DIFF
--- a/package-overrides/github/i18next/i18next@1.7.7.json
+++ b/package-overrides/github/i18next/i18next@1.7.7.json
@@ -1,0 +1,3 @@
+{
+  "format": "global"
+}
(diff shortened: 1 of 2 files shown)


guybedford:
Thanks, does this GitHub repo make the most sense to include then? What about the version on npm?

jerrytk:
I am pretty new at this stuff and so far I haven't been able to come up with working override for npm version :)
In my (maybe naive) opinion npm version pulls in a lot of other dependencies because of some added server-side i18next functionality which might not be all that useful in browser and would just add some overhead. It is basically a different repo: https://github.com/i18next/i18next-node

guybedford:
@jerrytk have you tried `jspm install npm:i18next-client`?

jerrytk:
Thank you for pointing that out! I have completely overlooked that package :( Client version works fine, I'll send new PR.

PR #118, opened 2015-01-20
Add bootstrap-bower@0.12.0.json

douglasduteil:
I'm not really sure if I have to use `"format": "global"`or not...

### DIFF
--- a/package-overrides/github/angular-ui/bootstrap-bower@0.12.0.json
+++ b/package-overrides/github/angular-ui/bootstrap-bower@0.12.0.json
@@ -1,0 +1,14 @@
+{
+  "dependencies": {
+    "angular": "~1.2.0"
+  },
+  "main": "ui-bootstrap-tpls",
+  "registry": "jspm",
+  "shim": {
+    "angular-bootstrap": {
+      "deps": [
+        "angular"
+      ]
+    }
+  }
+}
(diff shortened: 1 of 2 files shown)


douglasduteil:
@guybedford Do you have a test suite ?

guybedford:
@douglasduteil I have considered in the past if it is worth having all registry overrides come with tests, like the tests in https://github.com/jspm/jspm-cli/tree/master/testlibs.

But as the registry grows I think this may be unmanageable.

It is certainly worth considering to avoid regressions though.

guybedford:
Shim implies global format, so that works fine.

package-overrides/github/angular-ui/bootstrap-bower@0.12.0.json
```diff
@@ -0,0 +1,14 @@
+{
+  "dependencies": {
+    "angular": ">=1 <1.3.0"
```
guybedford: This isn't a valid dependency form for jspm.

guybedford: Either use ~, ^ or neither. Patches can be optional too.

douglasduteil:
Hum... I took this semver from [them repo](https://github.com/angular-ui/bootstrap-bower/blob/0.12.0/bower.json#L15).
Why can't we specify those kind of range ? 
Is the registery made fo store only minimal requirement ?
In those cases what should I prevent ? minimal versions (no version below 1.x) or maximal versions (no versions above 1.2.x ) ?

guybedford:
I would suggest setting to `~1.2` in this case.

douglasduteil:
Alright fixed

guybedford:
Great.

PR #121, opened 2015-01-20
Add ng-sortable@1.1.9.json

douglasduteil:


package-overrides/github/a5hik/ng-sortable@1.1.9.json
```diff
@@ -0,0 +1,17 @@
...
+  },
+  "directories": {
+    "lib": "dist"
+  },
+  "main": "ng-sortable",
+  "registry": "jspm",
+  "shim": {
+    "ng-sortable": {
+      "deps": [
+        "./ng-sortable.min.css!"
```
douglasduteil: This is to reproduce the css + js dependency here https://github.com/a5hik/ng-sortable/blob/1.1.9/bower.json#L8-L11

douglasduteil: Might be better to change the [main parsing](https://github.com/jspm/jspm-cli/blob/0.10.6/lib/package.js#L659) to always trust the file extension

So 
```js
"main": "ng-sortable"
// >>> loads "ng-sortable" without extension
"main": "ng-sortable.js"
// >>> loads "ng-sortable.js"
"main": "ng-sortable.css"
// >>> loads "ng-sortable.css" with the css plugin
```
 What you think @guybedford

guybedford: You can set the main to a CSS file, but in this case we want the js and CSS together. I am considering allowing virtual modules for this to allow variations of with and without CSS included, but for now what you've done is exactly the best advised way of going about it.

PR #122, opened 2015-01-20
Add angular-ui-select@0.9.6.json

douglasduteil:


package-overrides/github/angular-ui/ui-select@0.9.6.json
```diff
@@ -0,0 +1,19 @@
...
+  "directories": {
+    "lib": "dist"
+  },
+  "main": "select",
+  "registry": "jspm",
+  "shim": {
+    "select": {
+      "deps": [
+        "angular",
+        "./select.css!"
```
douglasduteil: Same as https://github.com/jspm/registry/pull/121#discussion_r23232806

PR #123, opened 2015-01-21
fix: update the aurelia default loader repo name

EisenbergEffect:


andreasgrimm:
Doesn't this break backwards compatibilty?
On the other hand I don't know how widely aurelia is used yet.

What about just adding `aurelia-loader-default` and leave `aurelia-loader-systemjs` as it is?
As I'm not completely sure, I'd suggest to let @guybedford have a look on this.
That's ok for you?

EisenbergEffect:
It's not even announced yet. There are a few people using it but they are in direct contact with me.

andreasgrimm:
Then it should be fine I guess :)

EisenbergEffect:
Thank you much! I don't intend to do that again. I wanted to get it taken
care of before I announce it though ;)

On Wed, Jan 21, 2015 at 2:00 AM, Andreas Grimm <notifications@github.com>
wrote:

> Then it should be fine I guess :)
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/jspm/registry/pull/123#issuecomment-70793134>.
>



-- 
Rob Eisenberg,
President - Blue Spire
www.durandaljs.com

PR #125, opened 2015-01-21
Add angular-touch@1.3.0.json

douglasduteil:


guybedford:
@douglasduteil just to follow up on the versioning - since this override will apply for all bower-angular-touch < 2, should we not be setting `angular` to a `^1.3` version, or does it need to lock exactly with the angular version for each one, and we just create a new override each time?

It would be worth also pushing for having the meta info in the package.json of the original repo so that future releases don't require this process.

douglasduteil:
Right

PR #127, opened 2015-01-21
angular-ui-tree moved to an organization

tauren:
This package stopped working because it was moved to an organization. I thought jspm handled github redirects, but it was getting a 404 finding the js file for angular-ui-tree after `jspm install` was run on the project. Updated the registry to reflect the correct place and added a new package-override. 

Is there any reason to keep package-overrides/github/JimLiu, or should it be removed as well? I don't want to break things for anyone, but I'm not sure it does any good anymore.


guybedford:
@tauren, right the redirect was supported by the registry entry but not with the override which is the correct behaviour. Thanks.

PR #134, opened 2015-01-23
feat(nprogress): add updated package override

EisenbergEffect:
This updates the override config based on feedback from @guybedford in the gitter channel. I have tested this and it works correctly with `import * as nprogress from 'nprogress';` I'm also submitting a PR to the nprogress repo to add the override to their package.json file.


guybedford:
Thanks!

Note the fact that this works with `import *` is a bug, it should be `import nprogress from 'nprogress'`.

I've created a bug for this at https://github.com/systemjs/systemjs/issues/330.

PR #135, opened 2015-01-24
added happy -> github:henrikjoreteg/Happy.js

Aaike:
added Happy.js

i have one question/comment though.

Happy.js can be used with either jQuery or Zepto
I have added this override without any dependencies so the user can choose to use it with either one.
But the user will have to remember to install either jQuery or Zepto manually.

i tested the override it works fine but is this the best way to handle this situation or should i be doing something else to handle the OR situation ?

### DIFF
--- a/package-overrides/github/henrikjoreteg/Happy.js@v0.8.json
+++ b/package-overrides/github/henrikjoreteg/Happy.js@v0.8.json
@@ -1,6 +1,13 @@
 {
     "main": "./happy.js",
     "registry": "jspm",
+    "shim": {
+        "happy": {
+            "deps": [
+                "jquery"
+            ]
+        }
+    },
     "dependencies": {
         "jquery": "jquery@^2.1.1"
     }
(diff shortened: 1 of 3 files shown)


Aaike:
also added jquery-validation

guybedford:
@Aaike it may be best to just make jQuery a dependency for `Happy`? Then if users want to use Zepto, they can override this dependency. Let me know your thoughts.

Aaike:
Yes i think you are right , there are way more jquery users ... and if they install it it will just work right away. and users who want to use Zepto probably know what they are doing, and can change it it the config.
i've changed it now to have jQuery as the default dependency

Aaike:
Maybe it would be nice to be able to map aliases and have the installer know about this.

something like:
    map :{
        "jquery": "zepto"
    }

If the users sets that up then the module loader could basically supply `zepto` instead if `jquery` if `jquery` is requested.
And the installer will skip any installation of jquery on further installs.

just a thought...

it could for example also serve well to let lodash be the replacement library for underscore without having to change the require statements.

guybedford:
@Aaike yes this could be done through custom overrides, which we're looking at making easier with https://github.com/jspm/jspm-cli/issues/398.

PR #136, opened 2015-01-24
Add registry and package override for DataTables

tejens:


guybedford:
Amazing, thanks and sorry for the delay!

guybedford:
Note I had to add a shim in https://github.com/jspm/registry/commit/1c37bce2b6ac31fdef808cb62530515a3a2eb955 to fix the jQuery dependency.

guybedford:
I had to revert the override for two reasons:

1. It isn't defining an object, so needs a proper shim (`require('datatables`)` was an empty object).
2. It could be worth considering including the CSS as part of the dependency process with the css plugin as well.

Let me know your thoughts on the above and we can consider adding it back, but would rather make sure we can add an override that is stable than one that might not be right.

tejens:
@guybedford I took a look at how Bootstrap was handling the jQuery dependencies and have made the following changes in #139

```json
"shim": {
    "js/jquery.dataTables": {
      "imports": "jquery",
      "exports": "$"
    }
}
```

With the above, `require('datatables')` will now return the jQuery `init()`. However, it seems that you still have to run `jspm install jquery` in order for the `jquery` mapping to be added to `config.js`. Is this the intended behavior?

I also tried using the `map` property in the override hoping that the `jquery` mapping would be added to `config.js` automatically when installing `datatables` but that didn't work.

```json
"map": {
  "jquery": "github:components/jquery@1.11.2"
}
```

I think including the CSS dependencies is definitely worth considering since it would provide a complete installation. I've also submitted #485 which would add the `jspm` override to DataTable's `package.json`.

PR #138, opened 2015-01-25
added angular-strap and angular-messages

capaj:
also moved angular-sanitize to keep proper alphabetical ordering


capaj:
@guybedford also I wanted to take this opportunity to ask you- how does package-overrides work exactly?

How is jspm working with the version of the override? I assume that it always uses the the latest override it can find?
For example for angular I see 6 identical overrides versioned differently. Why? Is this from before when jspm only used the override when it matched the version perfectly?

andreasgrimm:
thank you. One question though. Did you try and install e.g. angular-messages simply via the npm endpoint?
For me `jspm install npm:angular-messages` worked just fine.

andreasgrimm:
@guybedford regarding the existing angular registry entries with their regarding overrides, what would you suggest, how we should handle this in the future?
Until npm packages where available for angular stuff, we added registry entries refering to the `git:` endpoint (of their bower repo) along with the needed overrides. Now as of AngularJS 1.3 as they provide npm packages, I wonder what would be the best practice here (use the `npm` endpoint from now on / still use the `git` endpoint even for newly added entries / update the existing entries to use npm from now on / don't do anything of these and just always type `jspm install npm: ...`).

(sorry if I missed something along these lines that was discussed in other issues already)

capaj:
@andreasgrimm good questions. I think it would be best to have npm as preferred endpoint, because npm has more structure than raw github. You can't publish a module there without a package.json for example.  
Of course this is on @guybedford to decide.

andreasgrimm:
Yes, would be good to hear what he thinks about this (as I was away from other discussions recently, I'm not sure what's the best practice at the moment).

If we repoint the angular stuff to npm, we should check that overrides still work .. if all of them are still required.

angular-material makes for a special case here, as there's no bower/npm package yet .. and it's dependency to it's css file (therefore the additional SystemJS css plugin dependency) might be some kind of special, too.

guybedford:
@capaj at the comment at the end of https://github.com/jspm/registry/blob/master/README.md#submitting-a-package-override, it mentions how the versions apply. Help making this clearer is welcome, as I've been asked this a number of times before.

@andreasgrimm I would say use the npm version as that is the most likely case for browserify apps, and hence we will reduce version conflicts that way making the canonical version the npm version.

The thing to check though is if there are a lot of extra dependencies that get downloaded with the npm version. If so, we should add any `ignore` folders with an override (or just specify exactly which files / folders to include with a `files` array). This will then result in test dependencies like `require('fs')` that are in files not used by the main code, not leading to `fs` being a dependency for example.

The other thing to check is if the version on npm is defining itself in a way that can be used by other dependents. The standard issue here is something like Q on npm will not define a global, but many jQuery-style libraries that use Q expect a global to be defined. Since Angular carries its own modules namespace I don't think we would have this issue though?

capaj:
@guybedford I tried installing angular with trough npm endpoint and that is not good-it installs bunch of unneeded npm libraries. Until this behaviour is somehow rectified, we are better off as we are now-installing angular from github. So could you merge the PR now?

guybedford:
Sure.

capaj:
@guybedford thanks!

andreasgrimm:
@capaj @guybedford Sorry for being late on this. Currently I'm working on an angular-jspm example project (https://github.com/ng-next/ng-next-example) (seed will follow). That completely took the last couple of days away.

(After that I'll take a look into moving angular registry entries towards npm.)

PR #139, opened 2015-01-26
Added the proper shim for jquery dependency

tejens:


guybedford:
@tejens thanks for looking into this. Is it worth considering having CSS as a dependency as well with the CSS plugin?

tejens:
@guybedford Yes, I think it's definitely worth it since it would create a complete install. I was being selfish and going for a more minimal approach by only including the js file but having a complete installation is more reasonable. I'll start looking into and I can submit another pull-request once I'm done. Thanks for being so responsive and providing input :)

guybedford:
@tejens that would be great.

michaeldjeffrey:
I was trying to configure datatables and tabletools last night. 

TableTools takes straight DataTables. So when datatables exports `$` it breaks. 
This is where I ask the newb question of, how should plugins behave in regards to exporting. 

Should they export the parent and anything else (TableTools) has to deal with that?
Or do they export themselves (`$.fn.dataTable`) and it will remain on the parent?

guybedford:
@michaeldjeffrey what do you mean TableTools takes DataTables? Is it expected as a dependency? If so, is it expecting the `$.fn.dataTable` object exactly? This sounds very much like a special case if so.

michaeldjeffrey:
It [expects](https://github.com/DataTables/TableTools/blob/master/js/dataTables.tableTools.js#L36) the `$.fn.dataTable`.

What would be the correct way to handle this?

guybedford:
@michaeldjeffrey right, so you have two options here:

1. Run TableTools as CommonJS (the default here), then ensure DataTables exports `$.fn.dataTable` as you say.
2. Run TableTools as a global (using a `shim` or `format: 'global'` override), so that it catches the https://github.com/DataTables/TableTools/blob/master/js/dataTables.tableTools.js#L3223 path and sets the value itself.

Whether to use the global or CommonJS (or AMD) version then actually comes down to which makes sense for other compatibility reasons. Generally go with the CommonJS variant unless there is good reason to use the others.

So I'd suggest trying (1) above unless something else affects things.

In terms of plugin best-practises - the plugin should export whatever is appropriate to the exact plugin, I don't think there are general rules as it is plugin-specific if the `$.fn.x` value is nicely usable on its own.

tejens:
@guybedford updated package override with css plugin dependency. I also removed `"imports": "jquery"` from the shim config because it doesn't seem necessary.

guybedford:
Great! Ideally this should also be added to the original package.json as well.

PR #142, opened 2015-01-28
Add override for react-bootstrap

EliRibble:
Apparently jspm doesn't quite handle fuzzier dependency declarations


guybedford:
Thanks so much for posting the override though.

PR #147, opened 2015-01-30
Add DataTables TableToolst to registry

michaeldjeffrey:
This PR depends on #139 to be completed.


guybedford:
Let me know if this works with #139 ok. Is there a PR to add this to the original package.json?

michaeldjeffrey:
I was using the options from #139 when I was testing. 
 #148 fixes a flub I made.
There's a PR to add a package.json to TableTools

PR #149, opened 2015-01-31
Change infamous to npm.

trusktr:


trusktr:
Thanks to directories.lib for npm packages.

guybedford:
What's the reason you're wanting to use npm over GitHub?

trusktr:
Npm is the definitive place for javascript libraries, so it has that nice stature? xD

But for technical reasons, no specific reason at all. Do you recommend GitHub over npm?

guybedford:
I'd recommend only publishing to npm if the code you write has all its dependencies on npm and is CommonJS (registry: npm). Otherwise publish jspm code to GitHub (registry: jspm).

trusktr:
army-knife has no deps, and all of infamous' deps are on npm at this time. But that being said, it's not clear to me why this limitation needs to exist. Why does fo packages with registry:npm need deps only from npm? Why can't they have deps from GitHub? Why not let the package maker deside where the package's deps come from? Surely that package makers knows in what way his package works best.

trusktr:
Nvm, you answered my questions at https://github.com/jspm/jspm-cli/issues/454#issuecomment-72359918. Makes total sense, otherwise the package won't work with browserify or webpack workflows which rely on npm for package management.

trusktr:
It was in the back of mind mind that you mentioned to [prefer npm](https://github.com/jspm/jspm-cli/issues/177#issuecomment-58630060), I stumbled on it again:

> Generally when a package is on npm, it is preferable to use the npm version as all these sorts of edge cases get handled automatically.

guybedford:
@trusktr npm is preferred for existing packages written specifically for npm. New packages are favoured to be published on github.

trusktr:
Npm is the Golden Standard in my eyes, the god of package registries. I might not have mastered it yet, but I intend for npm to be the primary way of consuming the things I make, with GitHub as merely the code source for development. Things might change, but that's how it is right now. I'd be willing to bet that most people in the JavaScript world also see it this way, as far as package management goes.

guybedford:
To clarify what I'd recommend:

* If you're publishing to npm, that means your library works in plain npm and Node without jspm. Looking at your readme it looks like this is the case. I would suggest developing the library in the first place entirely against npm and browserify and only testing jspm as a "target environment".
* If you're wanting to use jspm-specific features like conditional-loading, plugins etc. We aren't going to be backwards compatible with npm. In this case you're better off just publishing to GitHub and not providing npm support at all.

I'm documenting some of the above more clearly today with some luck. Happy to merge this for now as you're in the first case it seems.

trusktr:
Okay, thanks! I'll keep that in mind. I definitely am aiming for as much consumability as possible. Infamous is just relying on `directories.lib` to get `/src` out of the require/import path for jspm usage.

PR #150, opened 2015-02-01
override for angular-material@0.7.1

capaj:
no more hammer.js needed


guybedford:
@andreasgrimm would appreciate your feedback on this - feel free to merge if you agree.

capaj:
@guybedford I tried it on my local machine by removing "dep hammer" from the header. Worked allright.

djindjic:
works for me too

guybedford:
Ok thanks. Is it an optional dependency then?

capaj:
@guybedford no they dropped the dependency, developed an internal solutions for handling touch events

guybedford:
Ok that sounds fine then, thanks.

andreasgrimm:
yes, it isn't included in https://github.com/angular/bower-material/blob/v0.7.1/bower.json anymore.
I tested the current override and it worked fine.

@capaj thanks for the override.
I guess at the time they give their ok to use it in production environments, they'll have a dedicated npm package for angular-material as well (as they currently do for almost all the other angular packages). So bower presumably will become obsolete.

PR #151, opened 2015-02-02
Update famo.us override.

trusktr:
This is theoretically what should work (i think?), but the shim doesn't add

```
"deps src/core/famous.css!";
"deps famous-polyfills";
```

to any files. Why is that again?

### DIFF
--- a/package-overrides/github/Famous/famous@0.3.0.json
+++ b/package-overrides/github/Famous/famous@0.3.0.json
@@ -1,13 +1,14 @@
 {
-    "directories": {
-        "lib": "src"
-    },
-    "dependencies": {
-        "css": "^0.1.0"
-    },
-    "shim": {
-        "*/*": {
-            "deps": ["../core/famous.css!"]
-        }
+  "directories": {
+    "lib": "src"
+  },
+  "dependencies": {
+    "css": "^0.1.0",
+    "famous-polyfills": "github:Famous/polyfills"
+  },
+  "shim": {
+    "*": {
+      "deps": ["src/core/famous.css!", "famous-polyfills"]
     }
+  }
 }
--- a/package-overrides/github/Famous/famous@0.3.0.json
+++ b/package-overrides/github/Famous/famous@0.3.0.json
@@ -7,7 +7,37 @@
     "famous-polyfills": "github:Famous/polyfills"
   },
   "shim": {
-    "*": {
+    "core/*": {
+      "deps": ["src/core/famous.css!", "famous-polyfills"]
+    },
+    "events/*": {
+      "deps": ["src/core/famous.css!", "famous-polyfills"]
+    },
+    "inputs/*": {
+      "deps": ["src/core/famous.css!", "famous-polyfills"]
+    },
+    "math/*": {
+      "deps": ["src/core/famous.css!", "famous-polyfills"]
+    },
+    "modifiers/*": {
+      "deps": ["src/core/famous.css!", "famous-polyfills"]
+    },
+    "physics/*": {
+      "deps": ["src/core/famous.css!", "famous-polyfills"]
+    },
+    "surfaces/*": {
+      "deps": ["src/core/famous.css!", "famous-polyfills"]
+    },
+    "transitions/*": {
+      "deps": ["src/core/famous.css!", "famous-polyfills"]
+    },
+    "utilities/*": {
+      "deps": ["src/core/famous.css!", "famous-polyfills"]
+    },
+    "views/*": {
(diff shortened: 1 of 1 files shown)


guybedford:
I think we came across this issue before - the shim wildcard doesn't work as a glob.

trusktr:
@guybedford https://www.npmjs.com/package/glob

guybedford:
@trusktr that is, for now, you need to directly shim the paths you are referring to - `"shim": { "some/path/*": ... }` will only catch the files in the specific folder at `some/path/...` and not deeper `some/path/.../...`.

trusktr:
Cool, let me try it and update the PR...

trusktr:
After trying what you suggested (see the new commit) then after installing famous none of the files contain `"deps src/core/famous.css!";` or `"deps famous-polyfills";`.

trusktr:
I can indeed verify that if I use a direct file name (without `.js`) then it works for that one file.

```
"deps src/core/famous.css!";
"deps famous-polyfills";
```

speigg:
Is there an explanation anywhere as to why the modules need to be modified at all? It seems like the override meta-data is duplicated. Why can't the override just be used directly by jspm when determining how to load a module?

trusktr:
@speigg It does work, but not with a `*` wildcard, not even a single wildcard at the end of a path. Direct paths to files work though. @guybedford Am I doing it wrong, or is this a bug?

guybedford:
@speigg good question. There's three options in terms of managing configuration of packages:

1. Inject all the package configuration into the global `config.js` file when you install a package. This will clutter the `config.js` file making it large, but we have a single source of truth for everything.
2. Have a `config.js` per-package that is loaded by SystemJS before loading the package itself. This way we keep the global `config.js` specific to its own needs, and have uncluttered fine-grained package config.
3. Inline config in the modules it applies to. The benefit of this approach is that config is no longer separate from packages - it is tied into their sources and doesn't clutter other areas or need a special request to load.

Currently we're using (3) although I'm still weighing up the merits of switching to (1) or (2).

speigg:
I think this is an important conversation to have (if it isn't being discussed anywhere already). Here is another option:

- Have a `config.js` for the main project configuration, and a `packages.config.js` For everything else (this way at least `config.js` is not cluttered). 

Also, another advantage of not doing (3), i.e, not modifying the original source files: doesn't that make it possible to symlink directly a linked package as part of the development workflow?

guybedford:
@speigg this is an important topic, I have left the architecture open to either method and the final decision point won't be for a few months yet, but it's important to work it out fully.

I like the suggestion, you're right about linking, although there are other operations like directories.lib and npm package builds which still need to be considered. Strangely linking is the one big argument for the single manifest. Note the CDN use case is something that pushes for in-file meta. Allways allowing either option is also an option though, and the loader supports this naturally as config can be a file or in-file.

trusktr:
@guybedford Sweet, so now these shims in package.json work:

```js
  "format":"amd",
  "jspm": {
    "registry":"jspm",
    "directories": {
      "lib": "src"
    },
    "dependencies": {
      "css": "^0.1.0",
      "famous-polyfills": "npm:famous-polyfills"
    },
    "shim": {
      "core/*": {
        "deps": ["src/core/famous.css!", "famous-polyfills"]
      },
      "events/*": {
        "deps": ["src/core/famous.css!", "famous-polyfills"]
      },
      "inputs/*": {
        "deps": ["src/core/famous.css!", "famous-polyfills"]
      },
      "math/*": {
        "deps": ["src/core/famous.css!", "famous-polyfills"]
      },
      "modifiers/*": {
        "deps": ["src/core/famous.css!", "famous-polyfills"]
      },
      "physics/*": {
        "deps": ["src/core/famous.css!", "famous-polyfills"]
      },
      "surfaces/*": {
        "deps": ["src/core/famous.css!", "famous-polyfills"]
      },
      "transitions/*": {
        "deps": ["src/core/famous.css!", "famous-polyfills"]
      },
      "utilities/*": {
        "deps": ["src/core/famous.css!", "famous-polyfills"]
      },
      "views/*": {
        "deps": ["src/core/famous.css!", "famous-polyfills"]
      },
      "widgets/*": {
        "deps": ["src/core/famous.css!", "famous-polyfills"]
      }
    }
```

So the files all have 

```js
"format global";
"deps src/core/famous.css!";
"deps famous-polyfills";
```

although but I was expecting to see `"format amd";`.

trusktr:
Of course `"deps": ["src/core/famous.css!"]` is wrong because `directories.lib` is already set to `src`.

PR #154, opened 2015-02-04
Add angular-translate

bargosc:


guybedford:
@andreasgrimm does this seem ok?

guybedford:
Merging for now, we can adjust if we need to.

PR #157, opened 2015-02-06
add override ionic-bower@1.0.0-beta.14.json

jimthedev:
this should be the last beta build before they release

### DIFF
--- a/package-overrides/github/driftyco/ionic-bower@1.0.0-beta.14.json
+++ b/package-overrides/github/driftyco/ionic-bower@1.0.0-beta.14.json
@@ -1,0 +1,23 @@
+{
+    "registry": "jspm",
+    "main": "js/ionic-angular",
+    "shim": {
+        "js/ionic-angular": {
+            "deps": [
+                "../css/ionic.css!",
+                "./ionic",
+                "angular",
+                "angular-animate",
+                "angular-sanitize",
+                "angular-ui-router"
+            ]
+        }
+    },
+    "dependencies": {
+        "css": "0.1.0",
+        "angular": "^1.3.6",
+        "angular-animate": "^1.3.6",
+        "angular-sanitize": "^1.3.6",
+        "angular-ui-router": "0.2.13"
+    }
+}


guybedford:
Awesome, thanks. Perhaps post a PR to see if they will take this configuration in their own package.json to avoid having to redo this?

jimthedev:
@guybedford. Will do. They've got a bunch of repos such as their ionicons and ngCordova so I'll see if I can work through them.

guybedford:
Thanks, that would be great! If you need any help just shout.

jimthedev:
@guybedford It looks like ionic does some funky builds where they use the main driftyco/ionic repository to generate a static version of ionic that then gets pushed as a release to the drifyco/ionic-bower repo. This means that I'll need to modify their build script to also copy a special bower.package.json from ionic to ionic-bower as package.json. All of this is fine, except I find myself wondering what to use in the bower.package.json that eventually becomes package.json in the bower repo.  

Would something like this work?

```json
{
  "name": "ionic-bower",
  "private": false,
  "version": "1.0.0-beta.14",
  "codename": "magnesium-mongoose",
  "repository": {
    "url": "git://github.com/driftyco/ionic-bower.git"
  },
  "licenses": [
    {
      "type": "MIT"
    }
  ],
  "jspm": {
    "name": "ionic",
    "version" : "1.0.0-beta.14",
    "main": "js/ionic-angular",
    "shim": {
        "js/ionic-angular": {
            "deps": [
                "../css/ionic.css!",
                "./ionic",
                "angular",
                "angular-animate",
                "angular-sanitize",
                "angular-ui-router"
            ]
        }
    },
    "dependencies": {
        "css": "0.1.0",
        "angular": "^1.3.6",
        "angular-animate": "^1.3.6",
        "angular-sanitize": "^1.3.6",
        "angular-ui-router": "0.2.13",
        "ionicons": "github:driftyco/ionicons@^2.0.1"
    }
  }

}

```

Note that the root 'name' property is ionic-bower whereas the jspm > name property is 'ionic'. I wasn't sure how to address this since the current shortname in the registry is ionic. Does it matter?

One other question is, do I need to include css as one of the dependencies since some css needs to be brought in for this to function?

guybedford:
This looks great to me - does it need all of those angular libraries? The CSS plugin should be a dependency, yes. Also how are `ionicons` loaded?

guybedford:
Also, no the name in the jspm package.json doesn't matter. The name and version in the package.json will only matter if and when we implement a `jspm publish`, but for now those are both ignored.

jimthedev:
Ionicons are currently loaded via a css link tag. They are used in some of ionic's base styles.

guybedford:
@jimthedev no pressure if it is tricky for the bower library process, but it's just if this makes maintenance easier.

jimthedev:
Shouldn't be too big of a deal. I'll work on it today.

PR #159, opened 2015-02-06
Added overrides and endpoint for three.js

BuildingJarl:
Example of how to use:

import THREE from 'three.js';
import 'three.js/examples/js/renderers/Projector';

### DIFF
--- a/package-overrides/github/mrdoob/three.js@r70.json
+++ b/package-overrides/github/mrdoob/three.js@r70.json
@@ -1,0 +1,10 @@
+{ 
+	"main": "build/three",
+ 	"ignore": ["docs", "src", "editor", "test", "utils", "examples/*.html", "examples/models", "examples/scenes", "examples/sounds", "examples/obj", "examples/textures"],
+ 	"format": "global", 
+ 	"shim": { 
+ 		"build/three": {
+ 			"exports": "THREE"
+ 		}
+ 	}
+ }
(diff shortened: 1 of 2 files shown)


guybedford:
Awesome, thank you.

Do you think you could send a PR to Three itself with this in the package.json? Happy to +1 such a thing if necessary.

BuildingJarl:
Sure no problem. just noticed a small mistake, but before I update I have a question: I added some ignores to the override. Should these be removed? As some people might need them.

guybedford:
Note that ignore can be a folder like `example` if you wanted to reduce this.

If it may be useful, yes you can remove them, but things like docs and examples etc don't need to be in the jspm version of the package.

BuildingJarl:
ok cool. #178 includes a small fix for the endpoint. Sorry to bother you but why again should this also be added to the three.js package.json? [Configuring-Packages-for-jspm](https://github.com/jspm/registry/wiki/Configuring-Packages-for-jspm) states that It can be added or left out. Since there is already an override, is it needed? Just Curious. Thanks

guybedford:
Because we need to add a new override every time a new version of Three is released. Having the jspm configuration in the package.json of the underlying repo is where it should really be in terms of maintenance and isolation. This package override service should be able to become obsolete eventually.

BuildingJarl:
Oh, I see. Makes sense. Thanks for taking the time to explain this.

PR #166, opened 2015-02-07
Add angular-http-auth registry and override

smajl:
Usefull angular module for intercepting, buffering and refiring HTTP requests.


guybedford:
What's the reason you've removed the other overrides here?

smajl:
Whoaaa?! LOL, those other commits were added automatically after the initial PR into it. Valid are only first three. Damn, OK, I will make another PR.

PR #169, opened 2015-02-08
add override for angular 1.3.12 and switch to unminified version

andreasgrimm:


guybedford:
Thanks! What do you think would be the chances of getting a PR for this into the Angular repo itself?

andreasgrimm:
I guess you mean the jspm property in angular's package.json itself? At least we could try.

andreasgrimm:
would this ease the switch for angular from github to npm as well?

guybedford:
Is this switch to npm something that we definitely want to do? Are there any obvious issues?

I guess we should work out that config first and do it in one PR instead of many separate ones to avoid churn.

andreasgrimm:
As far as I'm concerned using the github endpoint for angular is totally fine. I had no issues using them.

guybedford:
I think the only worry will be if the Bower repos stop being maintained, but that doesn't seem likely. So I'd be happy to continue with that too until there is good reason. In which case, it may well make sense to get jspm overrides in now then.

andreasgrimm:
sounds ok for me. I'll have to look when I'll be able to make some time for this.

guybedford:
Ok great, thanks so much, there's no real pressure actually since we've got these overrides for all angular1 versions anyway.

PR #170, opened 2015-02-08
Cleanup angular overrides and use unminified version of angular

smajl:
What I did here:
- Remove deprecated RC and beta versions
- Remove duplicated overrides covered by lower semver
- Add registries and fix angular-mocks ^1.2.0 version
- Use unminified version of angular library

I hope it's OK. :)


guybedford:
Note we do need to always keep overrides in as they are for backwards compatibility. This is important not to break existing libraries that have these dependencies still set.

Rather we should add new overrides for newer versions which provide the new behaviour where possible. If a new version isn't out yet, it can be good to try and align with the version release cycles of the packages out there.

It's a very different thing adding an override for a package not currently configured to work for jspm (meaning no one is relying on it), versus changing and adapting existing overrides, which needs to align with the natural versioning of the underlying package.

Let me know if you have any further questions about this.

smajl:
> The overrides will apply for all versions semver compatible with the override version, with the highest version override taking precedence.
Reading this I thought that we don't need several files with exact same override config `angular@1.2.x|y|z`, but we are good just with `angular@1.2.0` which covers everything. Am I wrong? :(

So when someone used `angular@1.2.18` override, he will be good with `angular@1.2.0` which has the same content.

guybedford:
Yes you're right, but prerelease versions don't fall under semver ranges - https://github.com/npm/node-semver#prerelease-tags.

smajl:
Oh I see, you mean these RC and beta version which I have removed. Well, I really doubt that someone is still using using angular@1.3-rc/beta when we have nice and stable 1.3 branch, but I get your point.

OK, so what now? Can you merge this with one eye closed? :D I promise I will we be very careful around any prerelease versions next time. :)

guybedford:
I'd suggest not removing those files, and making specific small changes. I'm also not sure I see the need for the `bower-angular-mocks` change. Adding the `registry: jspm` will not affect anything so that's fine for any override.

smajl:
I thought that the previous version of `bower-angular-mocks` was saying that: "if you need angular-mocks 1.2.7 and up, then you need angular 1.2.20 and up" - which seemed kinda strange to me. I wanted to correct that to target the whole 1.2.x line.

So if someone would like to install angular-mocks 1.2.6 and below via jspm then this override will still kick in?

Well, I have to close this then, angular.min was already corrected by someone else anyway. I will be more "atomic" next time. :) Thanks for your time.

guybedford:
Ok right - yes that sounds fine as a change on its own.

But yes do need to be more atomic as you say and careful with these changes. Merges push through to all users of jspm instantly, so it's not something to be done lightly.

smajl:
@guybedford @andreasgrimm Closing... The initial intention was this more arranged and clear.

Finally, I would suggest that contributors (e.g. @andreasgrimm) don't make unnecessary overrides for every minor version if it is exactly or almost the same config as in previous one. IMHO that should not be merged in first place.

In my opinion, it's better to have two `1.2.0` `1.3.0` files which covers everything (and add overrides only when something important changes) than ton of files for every minor version. Thank guys. :)

PR #175, opened 2015-02-08
Add angular-http-auth registry and override

smajl:
This time right... ;)

### DIFF
--- a/package-overrides/github/witoldsz/angular-http-auth@1.0.0.json
+++ b/package-overrides/github/witoldsz/angular-http-auth@1.0.0.json
@@ -4,7 +4,7 @@
   },
   "main": "http-auth-interceptor",
   "dependencies": {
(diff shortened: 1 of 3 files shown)


package-overrides/github/witoldsz/angular-http-auth@1.0.0.json
```diff
@@ -0,0 +1,15 @@
+{
+  "directories": {
+    "lib": "src"
+  },
+  "main": "http-auth-interceptor",
+  "dependencies": {
+    "angular": "^1.0.0 <1.2.0"
```
guybedford: Arbitrary version ranges are not supported unfortunately. Rather do `~1.1.0` perhaps.

guybedford:
Thanks, let me know when you've made the corrections.

smajl:
@guybedford Corrected.

PR #177, opened 2015-02-09
Update Zurb Foundation to 5.5.1

lyphtec:
SCSS files are needed in some Foundation workflows and should not be excluded.

### DIFF
--- a/package-overrides/github/zurb/bower-foundation@5.0.2.json
+++ b/package-overrides/github/zurb/bower-foundation@5.0.2.json
@@ -2,14 +2,12 @@
   "main": "js/foundation/foundation",
   "shim": {
     "js/foundation/foundation.*": "./foundation",
-    "js/foundation/foundation": ["jquery", "../vendor/custom.modernizr"]
+    "js/foundation/foundation": ["jquery", "../vendor/modernizr"]
   },
   "ignore": [
     "js/foundation.min.js", 
-    "scss", 
     "js/vendor/jquery.cookie.js", 
     "js/vendor/jquery.js",
-    "js/vendor/fastclick.js",
     "js/vendor/jquery.autocomplete.js",
     "js/vendor/placeholder.js"
   ],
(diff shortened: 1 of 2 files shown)


guybedford:
Nice, thanks. Can you explain the change for `vendor/custom.modernizr`?

lyphtec:
Hi,

That file no longer exists and has been replaced by just `modernizr.js` as per [this folder](https://github.com/zurb/bower-foundation/tree/master/js/vendor).

I've also included `fastclick.js` as this is handy to fix the 300ms click delay on mobile devices as recommended by Foundation [here](http://foundation.zurb.com/docs/javascript.html) since this is mobile first framework.

guybedford:
Ok right - perhaps do a new PR for a `@5.2.2` override then?

lyphtec:
You mean for the current version of `bower-foundation@5.5.1` ?

guybedford:
Right, yes exactly.

lyphtec:
Sorry, I'm still new to jspm - and haven't quite grasp how dependencies work.

What's your general preference on third party vendor references? Do you prefer the user explicitly specify it rather than have the override take care of it for them? Because currently it's a bit inconsistent as `jquery` is excluded... and so it's up to the user to add that to jspm.  
By the same token, I guess `modernizr` should be excluded too... and I can submit an override for `fastclick` to be added to the registry pointing to [this](https://github.com/ftlabs/fastclick)

guybedford:
jQuery is excluded because I was under the impression it is an optional dependency?

If not, we can add jQuery in as well, but yes - they should always point to the third-party versions for dependency-sharing.

lyphtec:
I'm pretty sure jQuery is required for most Foundation components to work as per [this document](http://foundation.zurb.com/docs/sass-files.html).. but `modernizr` & `fastclick` are optional.

So I'll make the necessary changes & will submit a new PR...

lyphtec:
Ok.. done.

I did a `jspm install` test with `-o` option and confirm that it looks ok as far as I can tell :smile:

guybedford:
Ok great thanks. Note we should still leave the old override in place, so if you can adjust this so that it just creates a new override file then I think we'll be good to go.

lyphtec:
Done

guybedford:
Great, thanks for getting this in!

PR #146, opened 2015-01-30
test: add registry tests

douglasduteil:
Following https://github.com/jspm/jspm-cli/issues/145
Here comes the registry tests 
result here : https://travis-ci.org/douglasduteil/registry/builds/48898010
Closes #119

PR #187, opened 2015-02-10
Joi

capaj:
added Joi object validation library and it's dependencies


registry.json
```diff
@@ -119,6 +120,7 @@
   "ionic": "github:driftyco/ionic-bower",
   "jcrop": "github:tapmodo/Jcrop",
   "jiff": "github:cujojs/jiff",
+  "joi": "github:capaj/joi-browser"
```
guybedford: Needs a comma.

registry.json
```diff
@@ -191,6 +193,7 @@
   "three.js": "github:mrdoob/three.js",
   "theseus": "npm:theseus",
   "theseus-angular": "npm:theseus-angular",
+  "topo": "github:capaj/topo-browser"
```
guybedford: Again, needs a comma at the end of the line.

capaj:
sorry, this just goes to show that it will be useful to have https://github.com/jspm/registry/pull/146

capaj:
@guybedford all comma separated now

guybedford:
Thanks!

capaj:
@guybedford  now when I install joi, I get only joi, not it's dependencies. I have to specify the dependecies under root, not under jspm right?

guybedford:
@capaj no the issue is it is installing 5.1.0 as the published tag. Rather try `jspm install joi@master` or tag a new stable version.

capaj:
@guybedford yes, silly me, forgot to tag new version. Thanks, almost works as expected-only it fails to match hoek version now:
```
capaj@capaj-i4771:~/WebstromProjects/tests$ jspm install joi
     Updating registry cache...
     Looking up github:capaj/joi-browser
     Downloading github:capaj/joi-browser@5.1.1
     Looking up github:capaj/topo-browser
     Looking up github:moment/moment
     Looking up github:capaj/hoek-browser
ok   Installed github:capaj/topo-browser@^1.0.3 (1.0.3)

warn No version match found for github:capaj/hoek-browser@2.x.x
ok   Installed github:capaj/hoek-browser@^2.12.0 (2.12.0)
ok   Installed github:moment/moment@^2.9.0 (2.9.0)

err  No version match found for github:capaj/hoek-browser@2.x.x

```

guybedford:
Yes `2.x.x` is not a valid version range in jspm, use `2`, `~2.0.0` or `^2.0.0` instead.

capaj:
@guybedford but my dependency is defined as "^2.12.0" https://github.com/capaj/joi-browser/blob/master/package.json#L19

guybedford:
So topo is using that https://github.com/hapijs/topo/blob/master/package.json#L16.

I'd advise installing from npm for these dependencies designed as npm modules.

capaj:
@guybedford of course, thanks for spotting that. I have tried installing them from npm, did not work out so well, that is why I created browser ports of these small libraries

guybedford:
@capaj great, do post issues if npm installs aren't working as expected.

capaj:
@guybedford will do

capaj:
@guybedford actually they use net, so 
```Uncaught Node net module not supported in browsers.``` is thrown.
Peculiar is that there are two ``` Uncaught RangeError: Maximum call stack size exceeded. ``` before that.
No point in raising this as an issue- Joi won't run unless they cut off the net module dependency as I did with Joi-browser fork.

PR #195, opened 2015-02-12
Add microplugin registry and override

smajl:
Final selectize dependency.


guybedford:
Thanks, apologies for the delay!

PR #196, opened 2015-02-12
changed mapping for less packages and added a less plugin

Aaike:
```
"less": "github:aaike/jspm-less-plugin",
"less.js": "github:less/less.js.js",
"less-dist": "github:distros/less.js",
```

`less` is the sytemjs plugin
`less.js` is the lib source library
`less-dist` is a distribution library containing minimal amount of files


guybedford:
Awesome, can we move the `less` to the top of the registry with the other plugins?

@matthewp do you mind if we change the `less` entry?

guybedford:
`https://github.com/distros/less.js` doesn't seem to exist?

guybedford:
I'd be interested to hear why we need both less.js and less-dist?

Aaike:
we do not , i'm just using less-dist , i just added the source lib to the registry for others

Aaike:
and originally i was using the source lib , until i decided to make the distribution repo :)

guybedford:
Ok right, and you're happy to maintain that? Would be great for others too, I think we can make that the primary version of less actually. Do we still need to create https://github.com/distros/less.js then?

Aaike:
https://github.com/distros/less is not working ?

Aaike:
i will not be updating every version of less in the distribution , but always happy to accepts PR's and makeing others admin , the distros namespace can be public

guybedford:
In the registry it's put down as `.js` - https://github.com/jspm/registry/pull/196/files#diff-d274c3c6a5ec7b84540c0a7a18abbcc8R119.

It's nice to have a primary repo for everything. Even if it's not up-to-date the fact that users can trust `require('less')` to work is the biggest thing. And then users can help you maintain it with PRs potentially as well. If you'd be happy to do this only though?

Aaike:
yes sure no problem , will maintain it as good as i can and hopefully other will help with time

Aaike:
i guess the github:less/less.js.js should also change to github:less/less.js :)

guybedford:
Ok great. Let's remove the other LESS and just make `less.js` the dist version?

matthewp:
@guybedford do not mind at all

guybedford:
Thanks, merged manually in https://github.com/jspm/registry/commit/e2697c70a1248bef1b37f3d1f54d2f99285b0e7b.

PR #197, opened 2015-02-13
Include jquery-ui

piuccio:
from github

### DIFF
--- a/registry.json
+++ b/registry.json
@@ -130,6 +130,7 @@
   "jquery-pjax": "github:defunkt/jquery-pjax",
   "jquery.cookie": "github:carhartl/jquery-cookie",
   "jquery.hammer": "github:hammerjs/jquery.hammer.js",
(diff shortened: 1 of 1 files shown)


guybedford:
Is this working correctly when loaded through jspm? Do we need to add a jQuery shim to the modules?

piuccio:
```
> jspm install github:jquery/jquery-ui
     Looking up github:jquery/jquery-ui
     Updating registry cache...
     Downloading github:jquery/jquery-ui@1.11.3

warn No main entry point detected for github:jquery/jquery-ui@1.11.3.

ok   Installed jquery-ui as github:jquery/jquery-ui@^1.11.3 (1.11.3)
ok   Install tree has no forks.

ok   Install complete.
```

piuccio:
I'm not sure if it works in the browser, I'm using `System.amdRequire`, not `import`

PR #202, opened 2015-02-14
New config for select2

smajl:
- select2 has now new official github repo select2/select2 (old ivaynberg/select2 is redirected to this)
- previous override did not have dependencies set

Is it OK to modify registry entry like this? If not, let me know.

### DIFF
--- a/package-overrides/github/select2/select2@3.4.5.json
+++ b/package-overrides/github/select2/select2@3.4.5.json
@@ -1,0 +1,11 @@
+{
+  "main": "select2",
+  "files": ["select2.js", "select2.png", "select2.css", "select2-spinner.gif"],
+  "shim": {
+    "select2": {
+      "imports": ["jquery", "./select2.css!"],
+      "exports": "$"
+    }
+  },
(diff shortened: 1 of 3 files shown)


guybedford:
Thanks. In order not to break the older versions of select2 we should only make the new override apply for the latest version, or at least the version since select2 stopped working. Then we should also copy the old package override to the new name as well.

smajl:
I can copy old override there, but even without it, it should not break anything I guess. I just added missing deps for jquery and css, therefore "repaired it". :) It's almost the same - new override only check, if you have jquery and css and jspm downloads them accordingly.

smajl:
@guybedford Copied old override. Let me know if I should make any other changes.

smajl:
@guybedford I copied the old override. Also, I've realized that it might come in handy to include locale files. I have used them before and they were not included in the previous override.

Hope it's OK now. Let me know, if anything else has to be changed.

guybedford:
Thanks, I've merged manually.

PR #204, opened 2015-02-16
adding bootstrap-datepicker

Jarreddebeer:


package-overrides/github/eternicode/bootstrap-datepicker@1.3.1.json
```diff
@@ -0,0 +1,18 @@
+{
+    "main": "js/bootstrap-datepicker",
+    "shim": {
+      "js/bootstrap-datepicker": {
+        "deps": [
+          "css",
```
guybedford: This line isn't necessary - rather add `css` as a dependency of the package to download - https://github.com/jspm/registry/blob/master/package-overrides/github/twbs/bootstrap%403.1.0.json#L5. Also is Bootstrap itself a dependency?

guybedford:
Thanks!

cAhmad:
Hi, they updated their project to match `npm` requierement.
The only override I needed to make the main lib works is: 
```json
"map": {
  "jquery": "@empty"
},
```
Because it brings `jQuery` from `npm` and I already had it from `Github`

But, importing a specific locale cause issue for me, I'm not able to make this works.
```js
import 'bootstrap-datepicker';
import 'bootstrap-datepicker/dist/locales/bootstrap-datepicker.fr.min';
```

When I bundle (_SFX_) my app it seems that the `locale` is called before the `dapicker` pluging is added to `jQuery` since I got this error: 

```log
bootstrap-datepicker.fr.min.js:2 Uncaught TypeError: Cannot read property 'dates' of undefined
```

I try to add in my `config.js`
```json
"meta": {
  "bootstrap-datepicker/dist/locales/bootstrap-datepicker.fr.min": {
    "deps": [
      "bootstrap-datepicker"
    ]
  },
}
```

It do not fix the issue. 
I have `jspm@0.16.36`, and it works almost (_somtimes it's broken like `0.16.36`_) correctly with `jspm@0.16.2`

Any help or tips would be really appreciated.

May relates to https://github.com/eternicode/bootstrap-datepicker/issues/1781 and https://github.com/jspm/jspm-cli/issues/1294

guybedford:
@cAhmad if you're able to share the above override with a new PR that would be great. Also the `map: { 'jquery': 'jquery' }` part shouldn't be necessary I think.

cAhmad:
Just submitted the PR. 
Yeah the `map: { 'jquery': 'jquery' }` is not necessary when using `jquery` from `npm`.

guybedford:
Great thanks for coming back to this.

PR #209, opened 2015-02-18
`three.js` should be `three`

speigg:


speigg:
Should `three.js` be deleted as well?

guybedford:
Thanks, I'm happy to include both for now. Unfortunately we can't really change registry names once added unless we modify jspm not to save names in the package.json like:

```json
{
  "dependencies": {
    "three.js": "r70"
  }
}
```

And rather store the full canonical name:

```json
{
  "dependencies": {
    "three.js": "github:mrdoob/three.js@r70"
  }
}
```

It's a minor aesthetic change that would allow the registry to change more regularly. Perhaps it's worthwhile.

speigg:
I think it would be worthwhile. It seems less mysterious that way.

guybedford:
Yeah I tend to agree. I've posted an issue at https://github.com/jspm/jspm-cli/issues/526.

PR #211, opened 2015-02-18
adding baconjs override

Jarreddebeer:
Set the name to "baconjs", as this occurs everywhere in their repository, but it's possible that "bacon" may be worthwhile.


guybedford:
Are you sure we need this override? I tried BaconJS without an override and it seemed to work ok for me?

smajl:
I would also suggest setting `main` to unminified files especially where sourcemaps aren't available.

Jarreddebeer:
I'm not sure the override is necessary, we could leave it out. Good point on the sourcemaps, i'll amend

PR #215, opened 2015-02-19
add vuejs registry.

wprater:
- the cjs modules are not linking correctly when being loaded with require(‘util’)
  where util/index.js exist, it’s trying to load util.js

### DIFF
--- a/package-overrides/github/yyx990803/vue@0.11.5.json
+++ b/package-overrides/github/yyx990803/vue@0.11.5.json
@@ -1,0 +1,3 @@
+{
+  "main": "dist/vue"
+}


guybedford:
Have you tried installing from npm - `npm:vue`?

wprater:
no, will give that a shot now. you think it'll fix the `require` issue as
well?

guybedford:
Yes the `npm` endpoint deals with the compatibility of the Node-style assumptions.

wprater:
ahh! can you point me to the code that deals with this? would love to
learn. will report back in a fee

guybedford:
It's not pretty, but it's at https://github.com/jspm/npm/blob/master/npm.js#L314 and https://github.com/jspm/npm/blob/master/npm.js#L414.

wprater:
installing from npm fixed it!

wprater:
is the general rule of thumb to load from npm if I know the package will be
using cjs module format?

guybedford:
Yes exactly - publish to npm for things npm supports. Publish to github for pure ES6 or use of jspm dependencies / plugins.

wprater:
I should setup a registry for `vue` to map to `npm:vue` then?  or is it
expected for users to know to try NPM for software using CJS?

guybedford:
Adding `vue: npm:vue` into the jspm registry is welcome.

PR #216, opened 2015-02-19
add vuejs registry.

wprater:


guybedford:
Note we don't need the override if it's empty. Merged manually.

wprater:
good to know. thanks.

PR #217, opened 2015-02-19
Added entry for angular-ui/ui-router for version 0.2.13.

jwstadler:
Currently, when executing the command `jspm install angular-ui-router`, the main file fails to be added properly. The addition of this registry entry fixes the problem. Here is the error message that appears in installation:

```
Updating registry cache...
Looking up github:angular-ui/ui-router
Downloading github:angular-ui/ui-router@0.2.13
Looking up github:angular/bower-angular

warn No main entry point detected for github:angular-ui/ui-router@0.2.13.

ok   Installed angular-ui-router as github:angular-ui/ui-router@^0.2.13 (0.2.13)
     Installed Forks

                           github:angular/bower-angular 1.2.28 1.3.13

     To inspect individual package constraints, use jspm inspect endpoint:name.

ok   Install complete.
```

### DIFF
--- a/package-overrides/github/angular-ui/ui-router@0.2.13.json
+++ b/package-overrides/github/angular-ui/ui-router@0.2.13.json
@@ -1,0 +1,11 @@
+{
+  "shim": {
+    "angular-ui-router": {
+      "deps": ["angular"]
+    }
+  },
+  "registry": "jspm",
+  "dependencies": {
+    "angular": "^1.3.2"
+  }
+}


guybedford:
Thanks so much for posting. I checked and there is actually already an override here - https://github.com/jspm/registry/blob/master/package-overrides/github/angular-ui/ui-router%400.2.12.json. I looked further and it turned out we had a regression stopping the main from working. I've posted a fix which should apply if you update jspm (which will update jspm-github here - https://github.com/jspm/github/commit/f7e6d8f6c80715eea9adc57b875355f57941d589).

Closing as this should be fixed now (note overrides apply for all future versions semver compatible).

jwstadler:
Sounds great. I will grab the latest version and give it a try. The behavior did seem a little odd since it worked previously. Thanks for checking into it. Have a great day!

PR #218, opened 2015-02-21
add plugin-typescript to jspm registry as 'ts'

frankwallis:


guybedford:
Amazing! Will test it out soon.

PR #222, opened 2015-02-23
famous pointing to famous-global.js

djindjic:


djindjic:
Current config of famous-angular override doesnt work.
I'm pretty sure this one in pr is correct configuration, but I'm little confused about already existing overrides without main section. Famous-angular package needs famous-global.js from famous package to make it works. This is based on [famous-angular documentation](https://github.com/Famous/famous-angular/tree/bebad621dc84e4cb3d817f36c19440f5483c18fa#inside-of-your-app). I have tested it with [this](http://codepen.io/zackbrown/pen/yyVQje) example and it works.
Also, I'm still waiting for confirmation of https://github.com/Famous/famous/issues/639, and if famous.js is just an amd wrapper over famous-global.js can we just use famous-global.js for famous jspm package as I set in this pull request (with format: "gobal")?

cc: @trusktr

trusktr:
@djindjic The problem with the global is you're automatically pulling in every single Famo.us class into your page, increasing download size and loading time. With famous.js (non-global) you're still doing the same unless you've got some type of pre-compiler to handle this special AMD file.

The ideal way to use famous would be to do `var Whatever = require('famous/whatever/Whatever')` or `import Whatever from 'famous/whatever/Whatever'` in your code, then have your build system import those modules as needed, and finally for production your compiled bundles would include only the classes/modules that you need.

The ideal, as @TheAlphaNerd would say, would be for famo.us to move to ES6 Modules, and for famous-angular to start using a module system instead of relying on the global.

The best course of action is to make a PR on famous-angular to make famous-angular *not* depend on the global file, and have it pull in dependencies using standard AMD, CommonJS, or even ES6 Module practices. I've started work on this in my fork of famous-angular, but haven't had time to get back to it.

If you change famous' registry override to "dist" instead of "src" in this PR, then libraries and applications that depend on the proper modular famous will fail to work as expected, and those devs will have to include "src" in every require/import as in `var Whatever = require('famous/src/whatever/Whatever')`.

A quick solution might be to create a new entry in the jspm registry called `famous-global` and let the override for that have `lib:dist`, then just `jspm install famous-global`.

trusktr:
@TheAlphaNerd @zackbrown It'd be super nice to move all code to ES6 Modules and Classes, which shouldn't be very hard at all since the prototype class pattern that Famo.us currently uses is essentially the exact pattern that ES6 Modules are syntactic sugar for. 

@zackbrown I started working on making famous-angular work with modules, but I haven't had the chance to get back to it. That'd be a nice step to take before migrating to ES6. Is there any plan to support Angular 2.0 (which seems like a dreadful migration)?

TheAlphaNerd:
In theory if you are doing smart dead code elimination requiring all of famo.us shouldn't be a problem.  This can easily be done with uglify or with closure compiler

TheAlphaNerd:
@trusktr we will not be moving to es6 modules or classes until they have landed with full support in all three vendors.  It was internally discussed for quite some time, but we cannot justify forcing transpilation on our end users

trusktr:
@TheAlphaNerd What about pre-publish transpilation? But yeah, it could wait. As for the issue of this PR it'd be great to get famous-angular using the current AMD modular system instead relying on the global. Then when famo.us is ready for ES6 that step will be out of the way.

trusktr:
cc: @jordanpapaleo ^ regarding famous-angular supporting famous AMD modules instead of the global.

TheAlphaNerd:
I heavily doubt that refactor would land in the angular integration.  There is no way to rely on generic AMD without including a modular loader in angular.

While optimizing not "loading in the entire platform" is a decent enough goal, I urge you to consider relying on external tools to handle dead code elimination as I mentioned before.  As for the inclusion of the package.

trusktr:
What do you mean by "including a modular loader in angular"? I don't think any loader (i.e. RequireJS?) is needed to be included. That would be up to the user and his build system. Jspm here, for example, loads AMD modules.

IMHO "dead code elimination" is a unnecessarily more complicated than simply using a module format. The best tools (browserify, webpack, jspm, etc) can simply load from any module format.

trusktr:
Here's a browserify-angular-commonjs example: https://blog.codecentric.de/en/2014/08/angularjs-browserify/

It could work out nicely! :beer:

TheAlphaNerd:
@trusktr this is creating the necessity for a build system.  Many developers use angular without any build system.  It is nice to enable these features, but you do not want to force it.

trusktr:
@TheAlphaNerd True! On that same token maybe we should avoid enforcing global usage so that it's not in the way of those who do choose a build system. famous-angular's package.json needs dependencies to complement bower.json and alternative build systems.

guybedford:
Is it worth considering using the version of Famous on npm?

TheAlphaNerd:
I was actually drafting a response earlier that it appears including the npm version should just work.

@guybedford what is considered the best practice for consuming a repo already published on npm?

guybedford:
@TheAlphaNerd the most advised workflow currently with jspm is to use npm for installing and publishing whenever possible.

guybedford:
Issues do typically arise though when you need the global defined. Famous angular probably expects to find the famous global?

We do have an issue to fix this in jspm by allowing globals to be defined from modules to ensure this layer doesn't make globals the lowest common denominator. This is being tracked at https://github.com/systemjs/systemjs/issues/148, I'm hoping to make some progress this week.

In the mean time the GitHub override may be preferable.

guybedford:
Is everyone happy to merge?

TheAlphaNerd:
:+1:

djindjic:
@guybedford, is it better to make famous-global override to service new famous-angular-global and leave current configuration for users whose don't to import all famous classes? It make sense because famouse-global.min.js is 194KB. On the other side, if users want just to `jspm install famous-angular` they cant make it works without some documentation which famous classes need to be imported separately.

guybedford:
@djindjic if users want modular famous, they can import individual modules as well instead of going through the main. But with famous-angular the only option is to load the whole thing surely?

djindjic:
as I've understood @trusktr imports just necessary famous modules:
>The ideal way to use famous would be to do var Whatever = require('famous/whatever/Whatever') or import Whatever from 'famous/whatever/Whatever' in your code, then have your build system import those modules as needed, and finally for production your compiled bundles would include only the classes/modules that you need.

this pr will brake his configuration, but famous-angular [readme](https://github.com/Famous/famous-angular/tree/bebad621dc84e4cb3d817f36c19440f5483c18fa#inside-of-your-app) says to import whole famouse-global in angular projects.

My vote goes to merge this pr, just want to check with you.

guybedford:
Why does this PR break that at all?

djindjic:
@trusktr :
>If you change famous' registry override to "dist" instead of "src" in this PR, then libraries and applications that depend on the proper modular famous will fail to work as expected, and those devs will have to include "src" in every require/import as in var Whatever = require('famous/src/whatever/Whatever').

guybedford:
Ahh right, I see. Would those alternative requires perhaps be possible for the npm version?

TheAlphaNerd:
If you want a modular version of famo.us can't you just install via npm?  There is no way to use the modular famous && famo.us angular.

There is no need to install via jspm if you just want common functionality

djindjic:
@TheAlphaNerd:
>There is no way to use the modular famous && famo.us angular.

I'm going to merge this pr. All developers depending on alternative requires need to use npm version.

thanks

guybedford:
Sounds sensible! Note that we are now bound to this API for the rest of the 0.3 branch.

djindjic:
one more thing.
does this [famous-angular](https://github.com/jspm/registry/blob/master/package-overrides/github/Famous/famous-angular%400.4.0.json) override, should allow just one line: `import 'famous-angular';`? I have tested this conf now and I need to `import 'famous';` manually before famous-angular, otherwise there is no famous defined.

guybedford:
Yes Famous needs to be added to the deps in https://github.com/jspm/registry/blob/master/package-overrides/github/Famous/famous-angular%400.4.0.json#L13.

trusktr:
I see this has already been merged, but I think there are a few better solutions to this. I'm willing to bet that large majority of people are using modular famous. This means now that the large majority has to do `jspm install famous=npm:famous`. I think it'd be better to satisfy the majority with the simple `jspm install famous` installing modular famous.

Possible solutions:

1. Instead make a `famous-global` entry in the jspm registry, and global users can use that.
1. Don't use `directories.lib` in this override, then when installing famous add a `paths` configuration so that `src` can be mapped out of require/import paths for modular famous users. The `main` entry can point to `dist/famous-global` for global users.
1. (the very best solution) Fix this problem in famous-angular instead of here and benefit not only jspm users but developers with all sorts of build systems.

TheAlphaNerd:
I think #1 is the best possible solution.  I want to quickly re-iterate that this will be a no-fix in famo.us angular.  It currently does not have any module loader and we are not looking to introduce one.

trusktr:
@guybedford Does an override support a `paths` (i.e. SystemJS paths)  configuration? If so, that'd make this problem much easier to solve, as described in possible solution 2. I know this can be configured in the application code, but in an override might be nice

trusktr:
@TheAlphaNerd 1 might be the best immediate solution, but it only benefits jspm users. Solution 3 would benefit a whole lot of people using different build systems.

trusktr:
@TheAlphaNerd 

> [famous-angular] currently does not have any module loader and we are not looking to introduce one.

I'm not sure what you mean. Why would it need a specific module loader?

trusktr:
@TheAlphaNerd There's a huge problem with famous-angular using the global famous: It makes developing addons to famous-angular very difficult to do in a modular way that can co-habit with vanilla modular famous. I tried for a solid evening, but it just doesn't work out, which is why I forked famous-angular and started converting it to a modular build.

famous-angular needs to use a module system in order to be compatible with modular famous for people who want to extend famous-angular in order to publish new fa-directives without being stuck having to use famous-global in their apps.

trusktr:
`angular.module` is one thing, but wrapping `angular.module` in a real module system is another thing, and IMHO is quite necessary and would satisfy many more developers, while using global famous would still be an option.

trusktr:
@TheAlphaNerd @zackbrown It's feasible to make files like this one (https://github.com/Famous/famous-angular/blob/master/src/scripts/directives/fa-canvas-surface.js) a CommonJS file, make a catalog file that `require`s the minimal set needed to set up famous-angular (easy to translate this to ES6 Modules later), then famous-angular users (the same ones using modular famous) can `require` the directives they need, and those directive modules (CommonJS modules) can `require` the services they need.

It'd work out very nicely for everyone as there can also exist a full-catalog file that requires everything for global users, or even better, there can simply be a new famous-angular-global repo for the global users.

trusktr:
@TheAlphaNerd, @derickbailey admits [here](http://derickbailey.com/2014/06/10/browserify-my-new-choice-for-modules-in-a-browser-backbone-app/) that Backbone.Marionette modules were a bad idea.

This is the same thing Angular has done, which is also a bad idea IMO. `angular.module` is basically a solution to JavaScript's lack of an official module system, but `angular.module` is limited to AngularJS projects. It's not ideal, which is why AngularJS 2.0 is dropping `angular.module` for ES6 Modules.

If there's not yet a plan to support Famo.us with Angular 2.0, then I'd say let's revert this commit and just make a `famous-global` entry to satisfy those users. It's the simplest solution.

TheAlphaNerd:
Whether or not they are a bad idea, they are what is supported by the general Angular ecosystem, and hence why it is what we support out of the box.  Despite being technically superior, your approach is unfortunately an edge case we cannot easily support.

If we were to revert this commit I would argue for just removing famo.us from the registry all together, as the npm version works as expected.

@trusktr if you wish to continue the discussion about refactoring FA I would suggest bringing the debate over there, at this point I feel like we are spamming the other people who are subscribed to this issue.

guybedford:
We can certainly consider making `npm:famous` the main registry entry for famous in jspm, since a user directly installing famous (without angular) would probably want that.

We recently shifted to make the registry only used at install time, and then the resolved package name is stored in the package.json so that we can change packages like this at any time. But we need to give deprecation time to that change.

I'd say in around 3 months we can make `famous` the `npm:famous`, but in the mean time will have to stick with things as they are.

PR #226, opened 2015-02-23
Added Angular Next to registry

robianmcd:


guybedford:
Nice work! Thanks for supporting jspm.

PR #228, opened 2015-02-24
Update famous@0.3.5.json

MajorBreakfast:
@djindjic I need direct access to the `src/` folder's files. I'm only including the physics engine and I'd like to use `jspm bundle` to only pack the modules I really need.

### DIFF
--- a/package-overrides/github/Famous/famous@0.3.5.json
+++ b/package-overrides/github/Famous/famous@0.3.5.json
@@ -1,15 +1,11 @@
 {
   "registry": "jspm",
-  "directories": {
-    "lib": "dist"
-  },
-  "main": "famous-global",
-  "format": "global",
+  "main": "dist/famous-global",
   "dependencies": {
     "css": "^0.1.6"
   },
   "shim": {
-    "famous-global": {
+    "dist/famous-global": {
       "deps": [
         "./famous.css!"
       ]


guybedford:
This is roughly backwards-compatible so can work actually. Note that the `"format": "global"` in the shim isn't necessary as shims are assumed global by default.

djindjic:
we could merge this with shim.format, it looks more declarative.
will this solve @trusktr's issue in https://github.com/jspm/registry/pull/222?

MajorBreakfast:
@djindjic No I think I take it out. After I think about it that makes totally sense: Why would one declare a shim if it was not a global module. After all, global modules are the only type of module that don't explicitly declare what the exports are.

MajorBreakfast:
@guybedford Is `"directories": { "rlib": "." }` necessary? I think it is because as far as I can remember it defaults to `dist/` if that is found. Ready to merge if yes, otherwise I'd like to remove it.

guybedford:
No, the directories.lib isn't necessary as there isn't already one in the Famous repo.

MajorBreakfast:
@guybedford I thought that jspm detects the `dist/` directory and uses that if present as root. So, that's not that case?
I removed it.

guybedford:
No it's not automatic - only if explicitly specified in directories in the package.json. Great thanks, merging.

guybedford:
Actually - the shim needs to be changed to `dist/famous-global` first I think?

MajorBreakfast:
@guybedford Oh yes :) Done

djindjic:
There is an error:
>warn No main entry point detected for github:Famous/famous@0.3.5.

(tried to fix it but I couldnt)

guybedford:
@djindjic thanks this was actually a bug in the GitHub endpoint - I've added a fix in 0.10.8.

MajorBreakfast:
@guybedford @djindjic Thanks for resolving this issue! (under a day :P)

djindjic:
@guybedford is it reference in jspm-cli and will be done after jspm-cli release publish?

PR #231, opened 2015-02-27
update(package-overrides): update shim and exports

ThomasBurleson:
Removed HammerJS dependency
Added angular-aria dependency
Added `exports` of 'ngMaterial' string.

### DIFF
--- a/package-overrides/github/angular/bower-material@0.8.2.json
+++ b/package-overrides/github/angular/bower-material@0.8.2.json
@@ -1,9 +1,8 @@
 {
   "main": "angular-material",
   "shim": {
-    "angular-material": {
-      "exports": "'ngMaterial'",
-      "deps": [
+    "exports": "'ngMaterial'",
+    "deps": [
         "angular",
         "angular-animate",
         "angular-aria"
@@ -15,8 +14,5 @@
     "angular": "^1.3.0",
     "angular-animate": "^1.3.0",
     "angular-aria": "^1.3.0"
-  },
-  "buildConfig": {
-    "minify": true
   }
 }


andreasgrimm:
Isn't the `./angular-material.css!` in the shim deps and the `css` SystemJS plugIn as a dependency needed anymore?

andreasgrimm:
.. on the other hand, I'd agree it'd make more sense to always import the CSS explicitly. Especially since the SystemJS CSS plugIn supports bundling now and would presumably require such an explicit import anyways.

Unfortunately I'm not entirely sure about the export of the 'ngMaterial' string. I'm guessing you can use this string in your angular app module dependencies .. and thus removing the need to memorize at least this one magic string (.. until other angular modules are changed to export their module name as a string as well)?

Thanks for fixing the `@master` override. Apparently it wasn't really used by anyone yet.

off topic: As angular material has made it to npm now as well (https://www.npmjs.com/package/angular-material), do you consider to add the jspm specific information directly to `package.json` in the future? This would make future overrides obsolete.

package-overrides/github/angular/bower-material@0.8.2.json
```diff
@@ -0,0 +1,18 @@
+{
+  "main": "angular-material",
+  "shim": {
+    "exports": "'ngMaterial'",
+    "deps": [
+        "angular",
+        "angular-animate",
+        "angular-aria"
+      ]
+    }
```
andreasgrimm: seems this curly brace should be removed.

package-overrides/github/angular/bower-material@0.8.2.json
```diff
@@ -0,0 +1,18 @@
+{
+  "main": "angular-material",
+  "shim": {
+    "exports": "'ngMaterial'",
```
andreasgrimm: seems the line with `"angular-material": {` is missing here

andreasgrimm:
testing the override (with the above mentioned missing line fixed), I couldn't get the export of the 'ngMaterial' string working though, neither by importing it as a regular export `import { ngMaterial } from 'angular-material';` nor by importing it as the default export `import ngMaterial from 'angular-material';`.
As I understand the jspm shim configuration's `exports`, the name of a global variable is expected here. 
Of course I might be missing out on something important here. I'm curious now what's the intended usage :)

Maybe @guybedford can take a look at this.

package-overrides/github/angular/bower-material@0.8.2.json
guybedford: Yes the module name needs to be the first property of shim, with the deps and exports object as its value.

guybedford:
@ThomasBurleson thanks for looking into this! @andreasgrimm you're right about the adjustments - would really appreciate your assistance with getting them tested.

For CSS, generally I do like including it, although it is up to the package author ultimately to decide.

Better than having this package info here is for it to be in the package.json of the original repo. These overrides are meant to be temporary until the package can include them itself.

andreasgrimm:
@ThomasBurleson regarding the CSS dependency, I'd be fine either way. But having it not included anymore, might break code that already relies on it. It shouldn't be a big problem though, as of it's still a zero point release and by you angular-material devs not labeled as production ready for quite some time (I guess).
Though, does anything speak against keeping it included from your point of view?
I could imagine scenarios where you'd not need to import the CSS, like when in your (controller or service) module/component you just need one of angular-material's services but not necessarily it's CSS. But that's just a guess and needs clarification.

@guybedford I'd be happy to help testing. As mentioned above, I already tested with the missing line fixed, which then worked .. except the export. That's something I'd be happy about if you and @ThomasBurleson could explain in a bit more detail how it's supposed to work and if my assumptions above are correct.

ThomasBurleson:
@guybedford, @andreasgrimm - I will respond with more notes/details later today. Thanks for your responsiveness. :+1:

PR #233, opened 2015-03-01
added jquery-ui dependency for fancytree

Aaike:
added missing jquery-ui dependencies for fancytree and jquery-ui-layout

### DIFF
--- a/package-overrides/github/allpro/layout@1.4.3.json
+++ b/package-overrides/github/allpro/layout@1.4.3.json
@@ -5,8 +5,8 @@
         "lib": "source/stable"
     },
     "dependencies": {
-        "github:components/jquery": "*",
-        "github:components/jqueryui": "*"
+        "jquery": "github:components/jquery",
+        "jquery-ui": "github:components/jqueryui"
     },
     "shim": {
         "jquery-ui-layout": {
@@ -14,3 +14,5 @@
         }
     }
 }
+
+
--- a/package-overrides/github/allpro/layout@1.4.3.json
+++ b/package-overrides/github/allpro/layout@1.4.3.json
@@ -9,7 +9,7 @@
     },
     "shim": {
         "jquery-ui-layout": {
-            "deps": "jquery"
+            "deps": ["jquery","jquery-ui"]
         }
     }
 }
--- a/package-overrides/github/allpro/layout@1.4.3.json
+++ b/package-overrides/github/allpro/layout@1.4.3.json
@@ -6,7 +6,7 @@
     },
     "dependencies": {
         "github:components/jquery": "*",
-        "github:components/jquery-ui": "*"
+        "github:components/jqueryui": "*"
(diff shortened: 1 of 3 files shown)


package-overrides/github/allpro/layout@1.4.3.json
```diff
@@ -9,7 +9,7 @@
     },
     "shim": {
         "jquery-ui-layout": {
-            "deps": "jquery"
+            "deps": ["jquery","jquery-ui"]
```
guybedford: We probably need to add jquery-ui as a dependency as well?

Aaike: I already added it last week or so : https://github.com/jspm/registry/blob/master/registry.json#L134

guybedford: We do need to explicitly state dependencies in the package.json as well as providing the shim config. The reason for this is that the `dependencies` ensures it is downloaded, while the shim just acts like a `require` statement. Note we also usually want to include full names in the package.json dependencies, like `github:components/jqueryui` the reason being that this way we can make backwards-incompatible changes to the registry itself without issue.

Aaike: right sorry, misunderstood you the first time, i totally forgot about adding it to dependencies property.
i did not know about the full names in the dependencies. will keep that in mind.

Aaike: is this correct now ?

package-overrides/github/allpro/layout@1.4.3.json
```diff
@@ -5,11 +5,12 @@
         "lib": "source/stable"
     },
     "dependencies": {
-        "jquery": "*"
+        "github:components/jquery": "*",
+        "github:components/jqueryui": "*"
```
guybedford: This should be:

```
"jquery": "github:components/jquery",
"jquery-ui": "github:components/jqueryui"
```

Aaike: just out of curiousity is this something new ? because most of the other overrides use something like :
```
"dependencies": {
        "angular": "^1.3.2",
        "angular-ui-router": "0.2.10"
    }
```

guybedford:
Great, thanks for this.

Aaike:
just out of curiousity is this something new ? because most of the other overrides use something like :
```
"dependencies": {
        "angular": "^1.3.2",
        "angular-ui-router": "0.2.10"
    }
```
so ideally the other overrides should also be changed ? my previous submissions were also like that

guybedford:
It was a relatively recent change - the idea is that usually we want to be explicit with dependencies in package.json files because the registry itself may change. Three.js may be one repo today, but another next year. And we don't want to be tied to ensuring `jspm install three` always has to resolve to the same thing.

PR #246, opened 2015-03-05
Added plugin-ember-hbs

OrKoN:
Is it ok if the name does not match the file extension? Ember hbs is not compatible with Handlebars but uses the same file extension (hbs).

### DIFF
--- a/registry.json
+++ b/registry.json
@@ -10,6 +10,7 @@
   "jsx": "github:floatdrop/plugin-jsx",
   "less": "github:aaike/jspm-less-plugin",
   "ts": "github:frankwallis/plugin-typescript",
+  "ember-hbs": "github:n-fuse/plugin-ember-hbs",
 
   "ace": "github:ajaxorg/ace-builds",
   "angular": "github:angular/bower-angular",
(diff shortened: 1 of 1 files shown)


guybedford:
Unfortunately not - users would need to do `import 'file.hbs!ember-hbs'` unless we add a feature to jspm to allow installs like `jspm install hbs=ember-hbs`.

guybedford:
Alternatively you can actually just do - `jspm install hbs=jspm:ember-hbs`.

OrKoN:
Running `jspm install hbs=jspm:ember-hbs` is fine by me. In package.json it will work like the following, right?

```
 "hbs": "jspm:ember-hbs@version"
```

But I don't want people to confused that this is a normal Handlebars plugin. Therefore I think ember-hbs is a good name

guybedford:
In package.json it should be - `hbs: 'github:n-fuse/plugin-ember-hbs` that is because the jspm registry can have names change at a later point in time so can't be relied on.

guybedford:
Ok great - will merge then. Can you just move it to the correct alphabetical position first, trying to stay organised!

OrKoN:
Updated. Thanks

PR #247, opened 2015-03-05
Added bootstrap-css-only@3.3.2

bargosc:


guybedford:
Is the goal here to save download time over `require('bootstrap/css/bootstrap.css!')`?

bargosc:
Sorry, didn't think of that.

PR #249, opened 2015-03-06
add ionic 1.0.0-rc.0 override

jimthedev:
Ionic 1.0.0-rc.0 is out. Here is an override.

### DIFF
--- a/package-overrides/github/driftyco/ionic-bower@1.0.0-rc.0.json
+++ b/package-overrides/github/driftyco/ionic-bower@1.0.0-rc.0.json
@@ -1,0 +1,23 @@
+{
+    "registry": "jspm",
+    "main": "js/ionic-angular",
+    "shim": {
+        "js/ionic-angular": {
+            "deps": [
+                "../css/ionic.css!",
+                "./ionic",
+                "angular",
+                "angular-animate",
+                "angular-sanitize",
+                "angular-ui-router"
+            ]
+        }
+    },
+    "dependencies": {
+        "css": "^0.1.5",
+        "angular": "^1.3.6",
+        "angular-animate": "^1.3.6",
+        "angular-sanitize": "^1.3.6",
+        "angular-ui-router": "0.2.13"
+    }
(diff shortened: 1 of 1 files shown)


guybedford:
Thanks. It could be worth semver-ranging the css or are you not a fan of the builder config that ensues?

Also, it may be worth looking into posting this override into the repo package.json itself for the main release if we can catch that.

jimthedev:
@guybedford 

Hmmm.. good q.

I'll have to look into the semver css ranging.  I'm not at all opposed to using it, even if it is verbose or requires a build out. They actually use scss and it is pretty common for people to compile their own, so I'm not sure how it would be best to handle this. There's two variations: one where people use the ionic bundle and another where sass is used with the ionic cli tool.  I will do some poking around to see if makes sense to semver the css. 

There is an outstanding but assigned PR for adding this into their build process that builds package.json is here: https://github.com/driftyco/ionic/pull/3061

package-overrides/github/driftyco/ionic-bower@1.0.0-rc.0.json
```diff
@@ -0,0 +1,23 @@
...
+                "./ionic",
+                "angular",
+                "angular-animate",
+                "angular-sanitize",
+                "angular-ui-router"
+            ]
+        }
+    },
+    "dependencies": {
+        "css": "0.1.0",
```
guybedford: I mean specifically to set this to `^0.1.5` etc.

guybedford:
Yeah I know it's hard to fit the existing patterns against the jspm patterns - happy to discuss through some options if it helps. Always best to try and get the override right as we are stuck with it once its in.

jimthedev:
@guybedford Good idea. I will try to bring up the pattern considerations on gitter so we can discuss. Added the semver range for the css dep. Thanks for the suggestions! I find it super helpful. Cheers again.

guybedford:
Great, looks good. Thanks.

Issue #203, opened 2015-02-16
Importing Superagent fails


psolbach:
Tried all kinds of variations, bumping requirements to same edge versions. No dice.
When issuing ```import Request from 'superagent';``` I get:

        Uncaught TypeError: Object prototype may only be an Object or null: undefined
        system.js:1 Uncaught Error evaluating http://localhost:8000/jspm_packages/npm/readable-stream@1.0.27-1/lib/_stream_readable.jsm@1.0.27-1/readable.js

What am I doing wrong here?

guybedford:
If you edit `readable-stream@1.0.27-1/lib/*.js` and manually replace `require('stream')` with `require('npm:stream-browserify@1.0.0/index')` that will get this to work.

In terms of a fix, I can add an override to make this work, but it won't be ideal as it involves locking down the version of stream. Will put some thought to a better solution.

guybedford:
Note there is a maximum stack sized reached error in the browser, although this is non-critical. Also working on a mid term fix for that.

psolbach:
Thanks for the quick reply. Now I'm getting a whole bunch of new errors w/ ```mime-types``` after the maximum stack errors.

    http:/localhost:8000/jspm_packages/npm/mime@1.2.11/types/mime.types
    index.js:9 Synchronous XMLHttpRequest on the main thread is deprecated because of its detrimental effects to the end user's experience. For more help, check http://xhr.spec.whatwg.org/.

and 

    GET http://localhost:8000/localhost:8000/jspm_packages/npm/mime@1.2.11/types/mime.types 404 (File not found) index.js:19

because the passed URL is missing the double slash.

guybedford:
@psolbach do you have a code example to cause that error?

psolbach:
But still the same ```import Request from 'superagent';```.
The call is in ```nodelibs-fs@0.0.1/index.js``` invoked by ```Mime.load()```.
Nothing installed except Vue, Superagent.
Stack-Trace:

```
exports.readFileSync (index.js:9)
Mime.load (mime.js:26)
(anonymous function) (mime.js:44)
(anonymous function) (mime.js:52)
(anonymous function) (mime.js:54)
doEval (system.src.js:2255)
__eval (system.src.js:2221)
a (system.src.js:335)
$__global.upgradeSystemLoader.e.instantiate.cjs.n.metadata.format.n.metadata.execute (system.src.js:1159)
f (system.src.js:667)
c (system.src.js:636)
[...]

PR #254, opened 2015-03-09
superagent@1.0.0

danharper:
Pretty sure I've got this right.

This works locally when running:

```
jspm install superagent=github:visionmedia/superagent@^1.0.0 -o x.json
```

I'm guessing this fixes this issue [**#203 - Importing Superagent fails**](https://github.com/jspm/registry/issues/203)?

---

Edit: I originally had a `dependencies` key in the object as the browser version of superagent only needs 2 dependencies, instead of all the ones in the node version. But apparently you can't override those.


guybedford:
Nice, thanks!

This seems to work.

Do you have any idea why `npm:superagent` hasn't been published at 1.0.0 yet? You'd think this would be published there as well.

danharper:
I'm not sure. It was only published 9 hours ago. Shall I hold off until they've tagged it on npm and switch to that instead of github?

guybedford:
Yeah that might be a good idea - makes more sense to use the npm published version when they do this.

danharper:
Cool, I'll keep an eye on it and ping you once it's tagged.

danharper:
Looks like the guy who's done the release just doesn't have npm access https://twitter.com/defunctzombie/status/574780655109734401, so won't be long.

danharper:
@guybedford superagent 1.0 is [on NPM now](https://www.npmjs.com/package/superagent), so I've updated this to point there instead of GitHub.

guybedford:
Thanks!

PR #257, opened 2015-03-09
Added an override for the most recent Ember.js Beta #5 (debug)

OrKoN:
I try to use the most recent version of Ember and it does not work w/o an override. There is a pull request to make the ember repo compatible with jspm (https://github.com/components/ember/pull/66) but it's not merged yet. In the mean time, I think it would be handy to have an override in the registry. 

A question in addition to the pull request: how do I switch between different versions of the lib? e.g. for development I would like to use ember.debug.js and for production ember.prod.js is better. And maybe the override should not include the debug version.. what do you think?

### DIFF
--- a/package-overrides/github/components/ember@1.11.0-beta.5
+++ b/package-overrides/github/components/ember@1.11.0-beta.5
@@ -1,0 +1,14 @@
+{
+  "main": "ember.debug",
+  "files": ["ember.debug.js"],
+  "dependencies": {
+    "jquery": "*"
+  },
+  "shim": {
+    "ember.debug": {
+      "deps": ["jquery"],
+      "exports": "Ember"
+    }
+  }
+}
+


guybedford:
The way to do this is just by requesting the different path - `require('ember/ember.debug')` vs `require('ember')`.

This can also be done with map config.

Should we not be including the other override info like in the link you sent though? Would be good to do this properly.

guybedford:
@MajorBreakfast would really value your feedback here.

MajorBreakfast:
I just created https://github.com/emberjs/ember.js/pull/10592. This should fix this. My other PR was never merged because it's a generated file.

@guybedford Mind looking over what I specified? Last time I made some mistakes.

OrKoN:
@MajorBreakfast :+1: thanks, if your PR gets merged, it would be just perfect.

OrKoN:
@MajorBreakfast Maybe let's define ember.prod as main?

OrKoN:
@guybedford the map config is an option. I always forget about it. Thanks. I think I will implement a gulp task that rewrites it for deployment.

Btw, I noticed that people tend to clear the map config in our setup (we install everything with npm install -> jspm install) if they have version conflicts because they think it's always automatically generated and not a self-standing thing.

MajorBreakfast:
> Maybe let's define ember.prod as main?

@OrKoN I chose the debug version because it displays valueable debug messages. It's more convenient to add the mapping from `ember.debug` to `ember.prod `in `config.js` just before you ship. As you say, it is sensible to automate this.

guybedford:
@MajorBreakfast thanks so much for pushing this through.

PR #258, opened 2015-03-10
Fix missing endpoits!

smajl:
Oh dear, no one noticed *facepalm* :D


guybedford:
Ha, thanks! Yes it would be good to have some automated registry tests for structure like this in due course.

PR #262, opened 2015-03-12
this breaks js-beautify

Aaike:
i commited this earlier , this was a mistake , sorry


guybedford:
An override file can't be empty. Rather delete it then?

Aaike:
This actually is a file deletion, it is no longer in the npm dir in my fork that's for sure

guybedford:
Ok sure, sorry was checking on my mobile phone and couldn't tell.

PR #264, opened 2015-03-12
add alias for angular-xeditable, fix override

andreasgrimm:
I know the changed override might break stuff. But the angular dep is missing and what's more important, it uses the minified version by default.

Would this still be ok though?

### DIFF
--- a/package-overrides/github/vitalets/angular-xeditable@0.1.8.json
+++ b/package-overrides/github/vitalets/angular-xeditable@0.1.8.json
@@ -1,3 +1,14 @@
 {
-  "main": "dist/js/xeditable.min"
+  "main": "dist/js/xeditable",
+  "shim": {
+    "dist/js/xeditable": {
+      "deps": [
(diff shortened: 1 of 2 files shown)


package-overrides/github/vitalets/angular-xeditable@0.1.8.json
```diff
@@ -1,3 +1,17 @@
 {
-  "main": "dist/js/xeditable.min"
-}
\ No newline at end of file
+  "directories": {
+    "lib": "dist"
```
guybedford: Let's leave this out here to maintain the structure since the override is already in use.

guybedford:
Yes changing the main and adding the shim is fine, as it is mostly backwards-compatible. Are there other files that can be required in the package? If so it's worth not alterning the directories.lib.

andreasgrimm:
ok. changed and rebased it to not alter the (default, inherent) directories.lib.

Issue #259, opened 2015-03-11
Registry tests


guybedford:
We should have a travis run on the registry that checks
* Registry.json is valid JSON
* registry.json has all names mapping to valid endpoints
* all package overrides are valid JSON

That initial start would be amazing, then from there we can grow to support more detailed package override linting.

guybedford:
Contributed by @2fd in https://github.com/jspm/registry/pull/273.

PR #273, opened 2015-03-19
Registry tests jspm/registry#259 + fix golden-layout@1.0.6.json

2fd:
command tool from file validation: `node test`

`node test --help`
image

`node test -e` display only errors if they occurred
image

image

`node test <package ...>` validate only `registry.json` and overrides from specific package
`node test ace angular react`
image

by default validates all
Example fail: https://travis-ci.org/2fd/registry/builds/54988938
Example passing: https://travis-ci.org/2fd/registry/builds/54989331

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 .DS_Store
+node_modules
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,10 @@
+language: node_js
+node_js:
+- '0.10'
+branches:
+  only:
+  - master
+notifications:
+  email:
+    recipients:
+    - i@2fd.me
--- a/package-overrides/npm/golden-layout@1.0.6.json
+++ b/package-overrides/npm/golden-layout@1.0.6.json
@@ -5,12 +5,12 @@
   "main": "dist/goldenlayout",
   "files": [
     "dist/goldenlayout.js",
-    "dist/goldenlayout.min.js",
+    "dist/goldenlayout.min.js"
   ],
   "format": "global",
   "registry": "jspm",
   "dependencies": {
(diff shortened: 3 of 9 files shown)


2fd:
* [x] Registry.json is valid JSON

* [x] registry.json has all names mapping to valid endpoints

> to validate the name of package uses the package Format property of "jspm/ npm" and "jspm/github"

* [x] all package overrides are valid JSON

> uses glob for find all override files of package

guybedford:
This looks amazing! Thank you so much for taking this on!!

I will aim to review this tomorrow for merging here, just need to finish the jspm release cycle first.

guybedford:
Ok, I'm merging this now because we need it! Then will continue to refine the code as we go.

guybedford:
Let me know if you're able to do further review work on this at all, otherwise will hack on it too when I have some time.

Thanks again so much for contributing!

2fd:
I would with pleasure.

Let me know anything else.

guybedford:
Ok, sure that would be amazing, just look at whatever you feel like - no pressure at all. The initial points would be just:

* Code style - the standard style we have is at https://github.com/jspm/jspm-cli/blob/master/.jshintrc, it would be nice to use this.
* OverrideFile - we seem to loop these from the registry. It is possible for something to be in the registry.json but not have an override and vice-versa. So it may be worth calling this "RegistryEntry" instead and making sure that sources are taken from both the registry.json and the directory listing. Alternatively we can treat the testing of the registry.json validation and testing of the override files as completely separate steps and remove the coupling between the two.

Let me know if that makes sense.

2fd:
yes! this make sense.

I now work in this

PR #278, opened 2015-03-19
Fixed export for foundation

bezoerb:
`import Foundation from 'foundation';` now sets the proper foundation export


bezoerb:
\o/

PR #286, opened 2015-03-21
Added Knockout-Validation

carlosagsmendes:


guybedford:
You probably need to add a shim here as well to the knockout-validation js file to make it dependent on knockout explicitly. Things can work otherwise, but will hit random bugs if knockout wasn't explicitly loaded first.

carlosagsmendes:
You're right. I've also added a main entry. Thks

guybedford:
Perfect, thanks.

guybedford:
Actually sorry I missed that the override name is case-sensitive against the repo so this still won't apply unfortunately until we fix the name - `package-overrides/github/knockout-validation/knockout-validation@2.0.2.json` to have the right case.

guybedford:
I manually did the rename in https://github.com/jspm/registry/commit/b825142f831ab72dcf4da1121120fb15569a905a. It should be working now.

carlosagsmendes:
Thank you, I tested it from scratch and it works fine.

One more question... In the knockout package override we have the following:

```
 "shim": {
    "dist/knockout": {
      "exports": "ko"
    }
  }
```

Where does this shim gets persisted when we install the package from jsmp?

I can't find anything about the exports in the config.js file.

guybedford:
Great, glad to hear that. It gets written into the file itself as the inline meta syntax - `"exports ko"`.

carlosagsmendes:
Got it! Thanks again

PR #287, opened 2015-03-21
added css-element-queries

Aaike:


guybedford:
We should try and fix to a release tag. I know its out of date though - any way to convince the author to tag? And perhaps include this override at the same time?

Aaike:
there is a tag actually , but the tagged version is buggy , if i create an override for the tagged version will it also work for the master ?

guybedford:
No it won't - can you get them to tag a 0.0.3?

Aaike:
i will try , will submit a 0.0.2 for now

Aaike:
Hi , i have asked them to tag a new release, 
however... for some reason they tagged it as "v0.2" instead of '0.0.3".
So i should rename the override to `css-element-queries@v0.2.json` ?

guybedford:
Yes exactly. `v0.0.3` is detected as a semver usable in ranges, but `0.2` or `v0.2` will not be.

Aaike:
can i close this one ?

guybedford:
Sure, sorry don't get notifications for added commits.

PR #290, opened 2015-03-23
new test

2fd:
Features:

- `node test` command now validates the registry.json and all overrides files.
    - registry.json is validated as before
    - overrides files:
        - the name is validated according to their endpoint
        - must respect the format `{NAME}@{VERSION|BRANCH}.json`
        - must be valid json
- `-e` flag only show the error messages (as before)
- test of packages specific `node test <package ...>`
    - if `package` is a key in registry.json, only is validated the format and overrides files for that key
    - if `package` is a package name, only is validated the overrides files for that package
    - As before, it is possible to test more than one package|key at a time.

image


test/lib/RegistryEntryCollection.js
```diff
@@ -0,0 +1,53 @@
...
+
+var path = require('path');
+var glob = require('glob');
+var mout = require('mout');
+var forOwn = mout.object.forOwn;
+var FILE_EXTENSION = /[^.]*$/i;
+var PACKAGE_EXPECTATION = /(.*)@(.*).json$/;
+
+/**
+ * @constructor RegistryEntry
```
guybedford: Can we call this "Overrides.js"?

guybedford: Or perhaps OverrideList / OverrideFiles?

test/lib/RegistryFile.js
```diff
@@ -15,5 +17,32 @@ var RegistryFile = module.exports = function RegistryFile(){
 	JsonFile.call(this, path.resolve('./registry.json'));
 };
 
+util.inherits(RegistryFile, JsonFile);
```
guybedford: OverrideFile?

guybedford:
Thanks this is amazing! And has already caught some issues.

I'm just trying to get a clear picture of the classes, which would be easier if we can use a single terminology:

* An **override file** or **override** is a package.json override file.
* A **registry entry** is just the mapping `"jquery": "github:components/jquery"`
* The **registry entries** are represented in the `registry.json`.

I was a little confused by using `RegistryFile` to refer to `OverrideFile`. Let me know if that makes sense?

test/lib/RegistryEntryCollection.js
2fd: OverridesList seems more appropriate because the module does not represent a file itself is just a list of strings

2fd:
:ok_hand:

2fd:
the override file can only have a valid semver?

branch names are not accepted?

guybedford:
No, branch and tag names are allowed fine.

2fd:
great! and finally, as you call it to formats `github:components/jquery` and `components/jquery`?

guybedford:
This one isn't a clear terminology yet unfortunately - I currently use `canonical package name` or `full package name` and `package name`.

2fd:
added GLOSSARY.md to collect terminology

test/lib/RegistryFile.js
```diff
@@ -1,12 +1,13 @@
...
 var path = require('path');
 var JsonFile = require('./JsonFile');
+var CANONICAL_NAME_EXPECTATION = /^([^:\s]*):([^\s]*)$/;
 
 /**
  * @constructor RegistryFile
- */		
-
+ */
 var RegistryFile = module.exports = function RegistryFile(){
```
guybedford: It would still be worth calling this OverrideFile instead of RegistryFile.

2fd: :ok_hand:  no problem

guybedford:
Great, thanks so much! Ready to merge?

2fd:
yes!

you might want to change the notification settings for travis.

https://github.com/jspm/registry/blob/master/.travis.yml#L10
http://docs.travis-ci.com/user/notifications/#Email-notifications

PR #291, opened 2015-03-23
fix(vertxbus): fix override for vert.x event bus bridge client again

frankwallis:


guybedford:
Can we perhaps use the `v2.1` tag here instead of master? It's not valid semver so we do need the `v`.

frankwallis:
done, I wasn't sure how to handle the invalid semver.

guybedford:
Thanks, yeah it's not ideal I know, but trying to use predictable rules.

Issue #296, opened 2015-03-26
validation for override properties?


2fd:
This is the list of all the properties that are used in at least an override file

```
directories
main
registry
dependencies
shim
files
format
buildConfig
ignore
map

dependencyShim
dependencyMap
directory
browser
jspmNodeConversion
```

I guess the last 5 are wrong.

You can add a key check and let override only the functional properties of npm:

```
files
main
bin
man
directories
repository
scripts
config
dependencies
devDependencies
peerDependencies
bundledDependencies
optionalDependencies
engines
engineStrict
os
cpu
preferGlobal
private
publishConfig
```

and those described here: https://github.com/jspm/jspm-cli/blob/0.14.0/lib/config/package.js#L40-L55

```
main
registry
dependencies
configFile
baseURL
lib
dist
packages
format
map
buildConfig
```

do you think?

guybedford:
This would be great!

The short list of allowed properties would be:

```
main
registry
dependencies
devDependencies
directories.lib
directories.dist
format
map
jspmNodeConversion
files
ignore
shim
browser
```

The legacy properties can be removed from old overrides:

```
directory
buildConfig
dependencyShim
dependencyMap
```

2fd:
removed!?

it not be better adapt dependencyShim (to shim) and dependencyMap (to map)?

guybedford:
They aren't working, so nothing breaks by removing them as this was 1.5 years ago.

We can certainly convert them though too.

PR #301, opened 2015-03-27
update properties

2fd:
#296 just missing `guybedford/require-css`


package-overrides/github/blueimp/jQuery-File-Upload@9.3.0.json
```diff
@@ -1,16 +1,13 @@
 {
     "main": "jquery.fileupload",
-    "directory": {
+    "directories": {
```
guybedford: Rather remove this property as it will break the shim.

guybedford:
Thanks so much for this! Feel free to delete the `require-css` override as that is not supported.

Issue #153, opened 2015-02-04
auth0-js


volak:
auth0-js depends on zuul which has a commit id reference causing
`unable to resolve defunctzombie/stacktrace.js#07e7b95`

Need to override dependencies to:

"dependencies": {
    "Base64": "~0.1.3",
    "debug": "~1.0.4",
    "json-fallback": "0.0.1",
    "jsonp": "~0.0.4",
    "packageify": "^0.2.0",
    "qs": "git://github.com/jfromaniello/node-querystring.git#fix_ie7_bug_with_arrays",
    "reqwest": "^1.1.4",
    "trim": "~0.0.1",
    "winchan": "^0.1.1",
    "xtend": "~2.1.1"
  },

guybedford:
thanks for letting me know, I added the override in https://github.com/jspm/registry/commit/82af02c819d5cd419b8648b1c420fc29c48c5d7e and it should be working now - let me know if you have further issues at all.

guybedford:
Note it is usually advisable to make a PR to the original repo to try and iron out the issue as well too.

PR #309, opened 2015-04-01
added package override for semantic-ui - dependency derekslife/wrench

amitos80:
Installing semantic-ui using: 
```
jspm install npm:semantic-ui 
```
failed due to the following dependency:

```
git://github.com/derekslife/wrench-js.git#156eaceed68ed31ffe2a3ecfbcb2be6ed1417fb2  
```
*I noticed a similar problem here: https://github.com/jspm/registry/issues/153

this seems to be a reference to the latest commit so it was possible to remove the hashtag.


package-overrides/npm/semantic-ui@1.11.6.json
```diff
@@ -0,0 +1,5 @@
+{
+  "dependencies": {
+    "wrench"            : "git://github.com/derekslife/wrench-js.git"
```
guybedford: Is there not a version on npm itself you can use here?

amitos80: Hi, 
it seems that the official npm for wrench js is based on github.com/ryanmcgrath/wrench-js
while the official Semantic UI package (https://github.com/Semantic-Org/Semantic-UI/blob/master/package.json) Is based on a fork (derekslife/wrench-js), which has no npm.

guybedford: Ok sure!

PR #310, opened 2015-04-02
Added three.js@master

BuildingJarl:
Added three.js@master


guybedford:
We don't really do overrides on master when there are other versions available.

BuildingJarl:
Sure no problem

PR #312, opened 2015-04-05
change lodash to github version

Aaike:
The npm version does not work properly in the browser,

### DIFF
--- a/registry.json
+++ b/registry.json
@@ -181,7 +181,7 @@
   "less.js": "github:distros/less",
   "lightbox2": "github:lokesh/lightbox2",
   "localforage": "npm:localforage",
-  "lodash": "npm:lodash",
+  "lodash": "github:lodash/lodash",
   "marionette": "github:marionettejs/backbone.marionette",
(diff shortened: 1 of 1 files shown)


Aaike:
uh , not sure why travis is failing on this , can't see anything wrong

guybedford:
Specifically what issues is the npm version giving?

Aaike:
hmm i installed it earlier and i got an error that it couldn't find browser.js in the process package.
it gets installed like this:
```
"lodash": "npm:lodash@3.6.0",
"npm:lodash@3.6.0": {
    "process": "github:jspm/nodelibs-process@0.1.1"
}
```

So i installed the github version which seems to work and it didn't add the process nodelib.

anyway i just removed the github version and installed the node version , now it actually works without errors...
but i'm still curious about the nodelib process ? why would it pull that in for the browser ?

guybedford:
Lodash accesses the `process` global, so the browserify rule is to include the process shim to enable that.

PR #314, opened 2015-04-05
Creates font-awesome@4.3.0 override.

Zoltu:
This allows importing of the font-awesome CSS file without including the version/path like `import 'font-awesome'`.  Fixes https://github.com/systemjs/plugin-css/issues/26


guybedford:
Thanks for your patience here - I wanted to have time to test it.

We just need to add the CSS plugin itself as a dependency here.

I'll merge this in manually.

guybedford:
Added in https://github.com/jspm/registry/commit/655985cae364c93d8145c69b3d6eec58a955ed15.

PR #327, opened 2015-04-13
Package override for engine.io-client, xmlhttprequest dependency 

Avnerus:
This is to address: https://github.com/jspm/jspm-cli/issues/627
The referenced commit is the top commit in a branch called "add/ssl-support".
However, this fix depends on https://github.com/jspm/jspm-cli/pull/679

### DIFF
--- a/package-overrides/npm/engine.io-client@1.5.1.json
+++ b/package-overrides/npm/engine.io-client@1.5.1.json
@@ -1,6 +1,6 @@
 {
   "dependencies": {
-      "xmlhttprequest" : "git+https://github.com/rase-/node-XMLHttpRequest.git#add/ssl-support"
+      "xmlhttprequest" : "git+https://github.com/rase-/node-XMLHttpRequest.git#add/ssl-support",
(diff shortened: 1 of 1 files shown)


guybedford:
Thanks! This should be working on the next patch release then.

guybedford:
Note that the dependencies look like:

```json
  "dependencies": {
    "has-cors": "1.0.3",
    "ws": "0.4.31",
    "xmlhttprequest": "https://github.com/rase-/node-XMLHttpRequest/archive/a6b6f2.tar.gz",
    "component-emitter": "1.1.2",
    "indexof": "0.0.1",
    "engine.io-parser": "1.2.1",
    "debug": "1.0.4",
    "parseuri": "0.0.4",
    "parsejson": "0.0.1",
    "parseqs": "0.0.2",
    "component-inherit": "0.0.3"
  }
```

We should include all the other dependencies as well in the override otherwise they will be left out.

Avnerus:
Ah I see! OK Updated

PR #329, opened 2015-04-14
override for wordcloud2

Aaike:
forgot the override earlier...
ps: i asked them to tag a release , i will submit another PR then for the tagged version (if they tag it)


wtfguy:
+1

Issue #306, opened 2015-03-30
JSPM config for Semantic-ui


eliorcohen:
hey,

i was trying to create a definition file for   [semantic-ui](https://github.com/Semantic-Org/Semantic-UI "semantic-ui")   based on the bootstrap example.
i forked the semantic-ui and add the adjustments, but it's seems that the jspm cli from some reason don't retrieve the updates i have made.

i used the chace-clear command and the -f flag without success..
i'm using version 0.15.1

this is the command i run to add the lib
```
jspm install github:eliorcohen/semantic-ui
```

i have added this code into the package.json file hoping it will work:
```
"main": "./dist/npm",
"jspm": {
    "main": "semantic",
    "shim": {
      "semantic": {
        "deps": "jquery",
        "exports": "$"
      }
    },
    "files": ["components", "themes"]
  }
```

thanks,
Elior

guybedford:
@eliorcohen note that jspm will install the latest stable tag version. You probably want to use `jspm install github:eliorcohen/semantic-ui@master` to get this to work.

Note you can get the same with:

`jspm install github:Semantic-Org/Semantic-UI -o "{main: 'semantic',shim:{semantic: {deps: 'jquery', exports: '$'}}}"`

Do send a PR here with the override if you get it working, it would be useful to have.

eliorcohen:
i ended up using 

```
jspm install github:semantic-org/semantic-ui -o "{ main:'./dist/semantic', shim: { 'semantic-org/semantic-ui' : { deps: ['jquery'], exports: '$' } } }"
```
this load the js file successfully, but i cant find a way to inject the related css files, i tried to add the 'files' attribute with the override flag, but it's throwing me a warning that the main entry point is not defined...

eliorcohen:
i have also noticed that the module don't export anything..
if i'm manually add the **"exports $";** deceleration in the js file (the **"format global";** is automatically inserted).

semantic.js file
```
/* */
"format global";
"exports $";
 /*
 * # Semantic UI - 1.11.6
```
app.js file
```
import {Router} from 'aurelia-router';
import bootstrap from 'bootstrap';
import semantic from 'semantic-org/semantic-ui';

export class App {
  static inject() { return [Router]; }
    constructor(router) {
      
      // this will return an empty object unless i will
      // manually inject the 'exports $'; to the js lib file.
      console.log('semantic:',semantic);

    this.router = router;
    this.router.configure(config => {
      config.title = 'Aurelia';
      config.map([
[...]

PR #332, opened 2015-04-16
Add Semantic-UI

guybedford:
As in https://github.com/jspm/registry/issues/306#issuecomment-93691910, I've tested this and it is working fine for me.

@glen-84 the exports are useful so that the import value of semanticUI is jquery itself - `import $ from 'semantic-ui'` which saves some typing. Also I had to make the CSS file loaded relatively.

Feedback welcome before merge.

### DIFF
--- a/package-overrides/github/Semantic-Org/Semantic-UI@1.11.8.json
+++ b/package-overrides/github/Semantic-Org/Semantic-UI@1.11.8.json
@@ -1,0 +1,19 @@
+{
+  "directories": {
+    "lib": "dist"
+  },
+  "main": "semantic",
+  "shim": {
+    "semantic": {
+      "deps": [
+        "jquery",
+        "./semantic.css!"
+      ],
+      "exports": "$"
+    }
+  },
+  "dependencies": {
+    "jquery": "github:components/jquery",
+    "css": "github:systemjs/plugin-css"
+  }
+}
--- a/registry.json
+++ b/registry.json
@@ -251,6 +251,7 @@
   "saairey-bootstrap-wizard": "github:saairey/bootstrap-wizard",
   "select2": "github:select2/select2",
   "selectize": "github:brianreavis/selectize.js",
+  "semantic-ui": "Semantic-Org/Semantic-UI",
   "scoped-polyfill": "github:PM5544/scoped-polyfill",
   "showdown": "github:showdownjs/showdown",
   "sifter": "github:brianreavis/sifter.js",


glen-84:
I think that you should maybe exclude the CSS, unless you know of some easy way to avoid the FOUC.

guybedford:
@glen-84 FOUC is a fact of life at this point, and bundling CSS avoids it. Adding a link tag in the HEAD referencing the CSS will cause the injection not to double up as the CSS plugin checks for that, so it's no problem using both methods.

glen-84:
How does bundling CSS avoid it? SUI is the only CSS file that I'm importing, and I still see the FOUC. 

Would this be different if it was a native loader instead of XMLHttpRequest?

Good point about the CSS plug-in. (it seems to be case-sensitive, I almost thought that it wasn't working)

guybedford:
There's an option to bundle into a `separateCSS` file which can then be included with a link tag in the head.

Alternatively if your server is aware of jspm, it can use the trace of the modules in a page to inject the right link tags ahead of time.

maggiben:
There have been some changes in semantic-ui since the release of version 2.0, is there any chance to fix the registry to accommodate the new version ?

fractalf:
Trying to get this working with the skeleton-es2016, but even though it looks like sementic-ui is loading (pretty shure it is loaded) it looks ugly (no css rules seems to appy)
This is what I did:

`
jspm install semantic-ui=github:Semantic-Org/Semantic-UI
`

Changed `import 'bootstrap';` to `import 'semantic-ui';` in main.js
Changed `<require from="bootstrap/css/bootstrap.css"></require>` to `<require from="semantic-ui/semantic.min.css"></require>` in app.html

Not working. I noticed in FF firebug while inspecting HTML -> Style that "bootstrap" is not even mentioned when boostrap loads, but when semantic-ui is loaded it seems to be double up.

fractalf:
Throwing in a screenshot here
semantic

fractalf:
Someone on the Aurelia chat (Thanx z-brooks) made me aware that I wasn't using semantic-ui correctly. I was under the impression that both bootstrap and semantic-ui styled basic tags by default (input, nav, etc), but it seems they each use different classes to style elements. So semantic-ui is loaded like I thought, but the templates are made for bootstrap and will not trigger the same styles as in semantic-ui.

Sorry for spamming the thread, but I explain it here instead of deleting the posts incase someone else missunderstand like I did :)

PR #335, opened 2015-04-17
Updated Angular Material dependencies

ThomasBurleson:


ThomasBurleson:
Apologies - the commit should be squashed with message == "Update Angular Material dependencies to use Angular 1.3.15"

ThomasBurleson:
When I run this is test mode using:

```terminal
jspm install angular-material=github:angular/bower-material@master -o "{ main: 'angular-material', shim: { 'angular-material': {  'deps': [ './angular-material.css\!', 'angular', 'angular-animate', 'angular-aria'  ] } }, 'registry': 'jspm', 'dependencies': { 'angular': '^1.3.15', 'angular-animate': '^1.3.15', 'angular-aria': '^1.3.15', 'css': '^0.1.0' }}"
```

The install works, but I only get this:

```js
System.config({
  "map": {
    "angular-material": "github:angular/bower-material@master",
    "css": "github:systemjs/plugin-css@0.1.9",
```

But I need this:

```js
System.config({
  "map": {
    "angular": "github:angular/bower-angular@1.3.15",
    "angular-animate": "github:angular/bower-angular-animate@1.3.15",
    "angular-aria": "github:angular/bower-angular-aria@1.3.15",

    "angular-material": "github:angular/bower-material@master",
    "css": "github:systemjs/plugin-css@0.1.9",
```

How can I get the Angular libraries mapped also?

ThomasBurleson:
### Question 2:

The current **bower-material@master.json** uses

```js
  "main": "angular-material",
  "shim": {
    "angular-material": {
      "deps": [
        "./angular-material.css!",
        "angular",
        "angular-animate",
        "angular-aria"
      ]
    }
  },
  "registry": "jspm",
```

This means - I assume - that `angular-material.js` will be used as the main entry point. This is not a cjs, amd, or es6 format... it is a module-wrapper es5.

So we also have an index.js in CJS format. Should I do this instead:

```js
  "main": "index",
  "exports": "'ngMaterial'",
  "shim": {
    "angular-material": {
      "deps": [
        "./angular-material.css!",
        "angular",
        "angular-animate",
        "angular-aria"
      ]
    }
  },
  "registry": "jspm",
```

guybedford:
Thanks for looking into this. There are a few points to note here, which I've bulleted below.

* We shouldn't be removing the existing overrides - this is because there is existing code that relies on these.
* Ideally we shouldn't be adding an override to the master version. This is because the master version is constantly changing so will be a moving target. Is there not a release version that you can use? If you really do need to shim the master version I would suggest keeping this as a local override only.
* Having deep (secondary) dependencies being surfaced like they were installed directly (primary dependencies) is a feature being tracked in https://github.com/jspm/jspm-cli/issues/490 for peer dependency support. For now manually installing everything you need is the way to go.
* Using the `index.js` looks like it would actually avoid the need for a shim as it contains all requires necessary to bootstrap the load. It may well be a good idea to reference that rather.

ThomasBurleson:
@guybedford - Your response turnaround times are wonderful. Let me update the PR based on your feedback. I will ping you when ready.

ThomasBurleson:
@guybedford - Changes ready:

*  removed the overrides for *master*
*  restored older overrides (deprecated IMO)
*  updated the format and entry point
*  added comment to [jspm/jspm-cli#490](https://github.com/jspm/jspm-cli/issues/490)

package-overrides/github/angular/bower-material@0.9.0-rc1.json
```diff
@@ -0,0 +1,11 @@
+{
+  "main": "index",
+  "format":"cjs",
+  "registry": "jspm",
+  "dependencies": {
+    "angular": "^1.3.15",
+    "angular-animate": "^1.3.15",
+    "angular-aria": "^1.3.15",
+    "css": "^0.1.0"
```
guybedford: These should all include the full dependency name like `"angular": "github:angular/bower-angular@^1.3.15"`.

package-overrides/github/angular/bower-material@0.9.0-rc1.json
```diff
@@ -0,0 +1,11 @@
+{
+  "main": "index",
```
guybedford: Perhaps it's still necessary to shim the CSS?

guybedford:
Thanks this is looking great.

package-overrides/github/angular/bower-material@0.9.0-rc1.json
ThomasBurleson: I thought about that... then disregarded. After consideration and your feedback, it does make sense to shim that one. Stand-by...

package-overrides/github/angular/bower-material@0.9.0-rc1.json
ThomasBurleson: I will fix this also.

package-overrides/github/angular/bower-material@0.9.0-rc1.json
ThomasBurleson: @guybedford - so what does this shim addition mean exactly? I realize that Angular Material application depend upon the `angular-material.css` and the CSS plugin. But this shim does not load the CSS... so what is the shim doing?

```json
"shim": {
    "angular-material": {
      "deps": [ "./angular-material.css!" ]
    }
  }
```

ThomasBurleson:
@guybedford - ping again. I think we are ready.

package-overrides/github/angular/bower-material@0.9.0-rc1.json
```diff
@@ -0,0 +1,16 @@
...
+  "format":"cjs",
+  "registry": "jspm",
+  "dependencies": {
+    "angular": "github:angular/bower-angular@^1.3.15",
+    "angular-animate": "github:angular/bower-angular-animate@^1.3.15",
+    "angular-aria": "github:angular/bower-angular-aria@^1.3.15",
+    "css": "github:systemjs/plugin-css@^0.1.9"
+  },
```
ThomasBurleson: @guybedford - so what does this shim addition mean exactly? I realize that an Angular Material application requires the `angular-material.css` to be loaded. And we can use the JSPM CSS plugin. But this shim does not load the CSS... so what is the shim doing?

```json
"shim": {
    "angular-material": {
      "deps": [ "./angular-material.css!" ]
    }
  }
```

package-overrides/github/angular/bower-material@0.9.0-rc1.json
```diff
@@ -0,0 +1,16 @@
...
+  "shim": {
+    "angular-material": {
+      "deps": [ "./angular-material.css!" ]
+    }
+  }
+}
```
ThomasBurleson: @guybedford - so what does this shim addition mean exactly? I realize that an Angular Material application requires the `angular-material.css` to be loaded. And we can use the JSPM CSS plugin. But this shim does not load the CSS... so what is the shim doing?

```json
"shim": {
    "angular-material": {
      "deps": [ "./angular-material.css!" ]
    }
  }
```

ThomasBurleson:
@guybedford - any thoughts on a timeframe for merging this PR? We will be release 0.9.0-rc2 on Monday (which I would like to have incorporated also). Should I submit an **override** file for that one now ?

guybedford:
@ThomasBurleson sorry for the delay. I didn't realise you were running the release there. Merging for now, but better than having this override here is to have this exact configuration in the package.json of the release package itself so we don't need to keep updating it for each release.

ThomasBurleson:
@guybedford -  See **Bower-Material** [package.json](https://github.com/angular/bower-material/blob/master/package.json#L21-L34)

> Already anticipated that need. Thanks for the confirmation. ;-)

guybedford:
@ThomasBurleson the shim exactly does cause the CSS to be loaded?

ThomasBurleson:
I obviously missed that feature. By shimming the CSS [see above] the Loader will actually automatically load the angular-material.css.

> This feature is not well documented, IMO.

I also do not like the Shim loading because it is **not** clear when the CSS loads... 
If I use explicit load process:

```js
// Load the Angular Material CSS associated with ngMaterial
// then load the main.css to provide overrides, etc.

import 'angular-material/angular-material.css!'
import 'assets/app.css!'

// Load Angular libraries

import angular from 'angular'
import material from 'angular-material'

// Load custom application modules

import main from 'app/main'
```

Then I know that the Angular Material CSS loads **before** the `app.css`. If I use the shim-auto-load... this is NOT the case:

screen shot 2015-04-19 at 12 07 34 pm

guybedford:
Yes CSS should be created based on modular principles over exact ordering - that is styling by using more specific class names in the application CSS.

jsteenkamp:
I like the specific selectors approach. It makes app more robust/modular. On larger projects it's nice to have a separate bootstrap and main app file. Bootstrap has 3rd-party framework imports (like Angular, Angular Material...) and the Angular Material CSS. The main app imports the bootstrap module first and then the application modules and CSS. Using this structure you are likely to still get the app CSS loading before the angular material CSS. So specificity is good.

screen shot 2015-04-20 at 8 27 29 am

PR #336, opened 2015-04-17
Restore alpha sequence in registry.json

probins:
Recent changes left these files in the wrong sequence. Logically AngularFire might be under firebase. I'm keeping angularFire for bc, but it might be better to just get rid of this duplicate entry.


guybedford:
Thanks!

PR #337, opened 2015-04-19
Add jquery-validation-unobtrusive to JSPM registry

glen-84:


package-overrides/github/aspnet/jquery-validation-unobtrusive@3.2.2.json
```diff
@@ -0,0 +1,14 @@
...
+  "shim": {
+    "jquery.validate.unobtrusive": {
+      "deps": [
+        "jquery-validation"
+      ],
+      "exports": "$"
+    }
+  },
+  "dependencies": {
+    "jquery-validation": "github:jzaefferer/jquery-validation"
```
guybedford: It could be worth adding a version range here?

glen-84: How do you specify `>=1.8` ? (see [here](https://github.com/aspnet/jquery-validation-unobtrusive/blob/master/bower.json))

guybedford: `^1.8` is all that is needed. Arbitrary greater than is dangerous.

PR #340, opened 2015-04-20
add materialize

delebash:
add materialize github registry entry


delebash:
not working correctly closing

PR #341, opened 2015-04-20
Add override for kasperlewau/angular-bind-notifier

kasperlewau:
Added dependency shim and endpoint mapping. 

Installed/imported the package with the following (for testing purposes): 
```js 
jspm install angular-bind-notifier=github:kasperlewau/angular-bind-notifier
import 'angular-bind-notifier';
```

I've read through the [package testing](https://github.com/jspm/registry/wiki/Configuring-Packages-for-jspm#testing-configuration) and the [package dependencies](https://github.com/jspm/registry/wiki/Configuring-Packages-for-jspm#package-dependencies) sections but I cannot make much sense out of it, specifically in terms of package dependencies. 

I've tested manual overrides (`-o`) with the given dependency list (in this case, angular) and they appear to not yield a different result, so I'm not 100% that the changes made in `package-overrides/github/kasperlewau/angular-bind-notifier@0.0.4.json` are in fact needed. 

Please do let me know if I'm at fault here and/or if there is a scenario where my notion that it doesn't matter can be debunked.

### DIFF
--- a/package-overrides/github/kasperlewau/angular-bind-notifier@0.0.4.json
+++ b/package-overrides/github/kasperlewau/angular-bind-notifier@0.0.4.json
@@ -1,0 +1,10 @@
+{
+  "dependencies": {
+    "angular": "^1.3.0"
+  },
+  "shim": {
+    "angular-bind-notifier": {
+      "deps": "angular"
+    }
+  }
+}
--- a/registry.json
+++ b/registry.json
@@ -18,6 +18,7 @@
   "angular-agility": "github:AngularAgility/AngularAgility",
   "angular-animate": "github:angular/bower-angular-animate",
   "angular-aria": "github:angular/bower-angular-aria",
+  "angular-bind-notifier": "github:kasperlewau/angular-bind-notifier",
   "angular-bootstrap": "github:angular-ui/bootstrap-bower",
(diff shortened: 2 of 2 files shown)


guybedford:
With Angular plugins, you can be lucky and have Angular always load first in many cases, so that the shim is not easily noticable. But you will suddenly hit a condition later on where the Angular dependency hasn't loaded yet (perhaps when bundling), and then it can be hard to track down.

Merging as this looks right to me.

kasperlewau:
Fair enough - I suppose I won't get to experience the few cases where this could be a problem now with the shim supplied and all, at least not in the case of this particular package.

Cheerios! : )

guybedford:
If you want to be sure that the shim is working, you can look at the installed package and you should see at the top of the `angular-bind-notifier.js` file - `"deps angular"`.

guybedford:
Just had a look and you probably want it to be `dist/angular-bind-notifier` in the shim!

kasperlewau:
Doh! I rebased and amended that commit so many times over I must've slipped up the last time. It should definitely be `dist/`. 

And the annotation at the top I did notice when `lodash` was part of my dependency chain and had that shimmed out with a manual override. Didn't pay attention to the angular aspect. 

Would you like me to push a separate PR for the correct shim path?

guybedford:
No problem, that would be great thanks.

PR #342, opened 2015-04-20
Correct the shim path for angular-bind-notifier.

kasperlewau:
Now points to dist/angular-bind-notifier as intended, correction of  a mistake introduced in #341. *sorry*.

poke @guybedford

### DIFF
--- a/package-overrides/github/kasperlewau/angular-bind-notifier@0.0.4.json
+++ b/package-overrides/github/kasperlewau/angular-bind-notifier@0.0.4.json
@@ -3,7 +3,7 @@
     "angular": "^1.3.0"
   },
   "shim": {
-    "angular-bind-notifier": {
+    "dist/angular-bind-notifier": {
       "deps": "angular"
     }
   }


guybedford:
Thanks!

kasperlewau:
yw!

guybedford:
Sorry, I probably should have checked this more carefully. Note that the package main is currently set to `dist/bindNotifier`. It is worth checking if this makes the most sense, or if it is worth changing to the `dist/angular-bind-notifier`.

guybedford:
Oh it's just an alias it seems, so nothing lost, but might be neater to reference the `angular-bind-notifier` directly?

kasperlewau:
With the main set to `dist/angular-bind-notifier` I had to resort to a manual override of `-o "{'main': 'dist/bindNotifier' }"`. I suppose that's the idea behind `-o` though and the JSPM mapping(s). 

With the current package main, I get a functioning import/require in node and system based projects. In the case of f.e. Bower, you'd reference the `angular-bind-notifier.js` file with a regular script tag - so that's a non issue as far as I'm concerned. 

Is this a case of flavour/best practices that have eluded me, or do you envision a scenario where this could potentially burn? (I guess you already answered that with the *nothing lost* remark..).

guybedford:
```
  jspm install angular-bind-notifier -o "{ main: 'dist/angular-bind-notifier' }"
```

Works fine for me.

The point is that `bindNotifier.js` is just a pointer module so unnecessary bloat.

It is cleaner to add the main override to that, but not necessary as it works both ways.

Let me know if that makes some sense.

kasperlewau:
I do believe i remember why I added the pointer module in the first place. To get rid of the need to run 
`import 'github:kasperlewau/angular-bind-notifier@0.0.1';` in my app files. Which was a case of missing the map in `registry.json`. 

With all that said I'm not entirely happy with the current situation of the dependency definition. If I leave it as-is, any app that imports `angular` prior to this component would get a warning in regards to attempts to load angular more than once. 

I tested: 
```
jspm install angular-bind-notifier -o "{ main: 'dist/angular-bind-notifier' }"
```
 in the current state and that works just fine - aside from the multiple invokations of angular. Something you've ever experienced with other packages before or am I doing something *terribly* wrong here? : )

guybedford:
Sorry I completely didn't notice that this is your package.

Are you sure Angular is being invoked twice? What makes you think that? Typically you need to do your own Angular bootstrap when installing with jspm as the scripts don't load synchronously in the page.

Note that for your own packages, you can include these override properties in the package.json and don't actually need to post anything to the registry here.

kasperlewau:
> Sorry I completely didn't notice that this is your package.

No worries :) 

>Are you sure Angular is being invoked twice? What makes you think that? Typically you need to do your own Angular bootstrap when installing with jspm as the scripts don't load synchronously in the page.

Angular is being invoked twice as per the output in the console with the shim-deps annotations in place. 
If I remove the following from `dist/angular-bind-notifier`: 
```js
/* */
"format global";
"deps angular";
```

Angular would only ever be invoked once. 

This is how I bootstrap the application itself: 
```js
import 'angular';
import 'angular-ui-router';
import 'ui-router-extras';
import 'angular-bind-notifier';
import 'npm:ocLazyLoad@0.6.3';

import { default as app } from './app';

angular.bootstrap(document, [ app.name ]);
```

Leaving the dependency shim annotation in the package and commenting out `import angular-bind-notifier` will silence the warnings about double invokation(or rather, double loading). 

> Note that for your own packages, you can include these override properties in the package.json and don't actually need to post anything to the registry here.

I should've thought of that prior to this PR I reckon, that would've saved us both some time : )

guybedford:
Are you perhaps installing two versions of Angular? If you run `jspm inspect -f` do you see that?

kasperlewau:
The output of that command is multiple versions of my own package (been a couple o' updates today), but no forks of Angular itself. I'll remove all other dependencies and ensure that multiple Angular versions is not the case.

kasperlewau:
Cache cleared, fresh installation and all appears to be well. I'm stumped, which is a good thing I suppose.

In any case - the bloaty pointer module is gone in the latest version (0.0.5), do I need to do some manual update to the registry file and/or delete the entry? Add another override for v0.0.5 / delete the 0.0.4 one?

guybedford:
Great to hear that. Feel free to send a PR to remove the registry entry if you're going with your own package config.

kasperlewau:
Will do! On my way home right now, I'll get to it in a wee bit.

kasperlewau:
Just to clarify; when you say entry, you are referring to the package override and not the registry.json mapping, or both?

guybedford:
Yes just the package override.

PR #344, opened 2015-04-20
Remove redundant package override.

kasperlewau:
Removes an unnecessary package override introduced by yours truly.

Ref: #341 #342 

poke @guybedford - truly sorry about the slight mess I caused in the commit history here.


guybedford:
No worries, glad its making some sense.

PR #347, opened 2015-04-21
Add override for Angular Material 0.9.0-rc2

ThomasBurleson:


guybedford:
Do we need this? Is the previous override overriding the values you've set? If so it may be worth just posting a blank override to clear it. We're working on refining this "blanking out" workflow when the repo takes over.

ThomasBurleson:
See issue #348

guybedford:
Sure I've answered there, let me know if you think you still need this, otherwise we can just make it empty to erase the previous overrides - `{}`.

ThomasBurleson:
@guybedford - here is the blank override. Per #348, I updated Angular Material's [package.json](https://github.com/angular/bower-material/blob/master/package.json).

> Would you mind testing all of this in a temp directory to see if `jspm install angular-material` now uses that **package.json** and installs properly? Thx in advance.

guybedford:
It seems to load ok for me. One thing that could be worth trying is setting the exports of `angular-material` in the shim to angular to save needing two separate import statements in user code. But that may be an unnecessary convenience.

A blank override isn't needed for master as it shouldn't have any overrides applying - the overrides only apply when the version being installed is semver-compatible with the override version.

PR #349, opened 2015-04-21
Added malihu-custom-scrollbar-plugin

Aaike:
tested and working


guybedford:
Note we're no longer encouraging setting `files`, `ignore` or `directories` for overrides.

Aaike:
@guybedford what would be the alternative then to specify which files should be installed ?

Aaike:
hmm i added the "format": "global" option to the override.
any idea why it would have cjs [here](https://github.jspm.io/malihu/malihu-custom-scrollbar-plugin@3.0.8/jquery.mCustomScrollbar.js) ?

guybedford:
@Aaike the override hadn't refreshed on the CDN yet. Try it again now?

PR #362, opened 2015-04-27
Create nsPopover@0.6.7.json

nkoterba:
Add jspm registry support for nsPopover: https://github.com/nohros/nsPopover

Can't seem to get less/ns-popover.less file included although it's listed in "files". Is there a filter on .less files?

### DIFF
--- a/package-overrides/github/nohros/nsPopover@0.6.7.json
+++ b/package-overrides/github/nohros/nsPopover@0.6.7.json
@@ -11,8 +11,6 @@
   },
   "format": "global",
   "shim": {
-    "src/nsPopover": {
-      "deps": ["angular"]
-    }
+    "src/nsPopover": ["angular"]
   }
 }
--- a/package-overrides/github/nohros/nsPopover@0.6.7.json
+++ b/package-overrides/github/nohros/nsPopover@0.6.7.json
@@ -1,10 +1,5 @@
 {
   "main": "src/nsPopover",
-  "files": [
-    "src/nsPopover.js",
-    "less/ns-popover.less",
-    "sass/ns-popover.scss"
-  ],
   "registry": "jspm",
   "dependencies": {
       "angular": "^1.2.9"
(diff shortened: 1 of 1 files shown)


package-overrides/github/nohros/nsPopover@0.6.7.json
```diff
@@ -0,0 +1,16 @@
+{
+  "main": "src/nsPopover",
+  "files": [
+    "src/nsPopover.js",
+    "less/ns-popover.less",
+    "sass/ns-popover.scss",
+  ],
```
guybedford: We try to avoid adding this property in overrides.

nkoterba: You mean the "files" property or the main one?

guybedford: Yes the `files` property.

nkoterba: Ok I can remove the 'files' property, however I was following sample overrides from the registry and many of them include the files override.

nkoterba:
Any idea why the less/ns-popover.less file is being ignored/not pulled down when I do a `jspm install github:nohros/nsPopover` with the current package registry override: https://github.com/nkoterba/registry/blob/df7bddb95d4f91f1adfce4f3cc8df274e0946f03/package-overrides/github/nohros/nsPopover%400.6.7.json

guybedford:
It doesn't seem to be there - https://github.com/nohros/nsPopover/tree/0.6.7/src.

nkoterba:
@guybedford Ugh...you're right...it's in #master but not in version #0.6.7. Odd.

I don't maintain that package so I will file an issue that they should either push out a newer version or update their tag.

Thanks for the help and feedback.  Hopefully my future jspm registry pull requests will go smoother now that I better understand the process.

PR #365, opened 2015-04-30
Updated registry for golden-layout to point to version 1.0.8 and github

nkoterba:
Left old npm/golden-layout@1.0.6.json file in npm overrides.  Needed version 1.0.8 of golden layout via jspm.  In addition, the npm/golden-layout@1.0.6.json file was using the frowned upon `files` entry, and not including things like the css source files, etc.

### DIFF
--- a/package-overrides/github/hoxton-one/golden-layout@1.0.8.json
+++ b/package-overrides/github/hoxton-one/golden-layout@1.0.8.json
@@ -1,0 +1,11 @@
+{
+  "main": "dist/goldenlayout",
+  "format": "global",
+  "registry": "jspm",
+  "dependencies": {
+    "jquery": "*"
+  },
+  "shim": {
+    "dist/goldenlayout": ["jquery"]
+  }
+}
--- a/registry.json
+++ b/registry.json
@@ -149,7 +149,7 @@
   "font-awesome": "npm:font-awesome",
   "foundation": "github:zurb/bower-foundation",
   "fs": "github:jspm/nodelibs-fs",
-  "golden-layout": "npm:golden-layout",
+  "golden-layout": “github:hoxton-one/golden-layout",
   "hammer": "github:hammerjs/hammer.js",
   "handlebars": "github:components/handlebars.js",
   "handlebars.js": "github:components/handlebars.js",


guybedford:
@nkoterba can you confirm what the reason is for switching to the GitHub version? The CSS files do seem to be in the `src/css` on npm.

The `files` and `directories` options are only to be avoided in added overrides, and aren't a problem if added by the package author in the package.json.

nkoterba:
@guybedford I was hoping the 'golden-layout' jspm and/or npm registry entry was correct.  However, when I run:
```javascript
jspm install golden-layout
```

This is what the console outputs:
```javascript
nk-mbp:webclient nkoterba$ jspm install golden-layout
     Updating registry cache...
     Looking up npm:golden-layout

warn Main entry point not found for npm:golden-layout@1.0.8.
     Adjust this property in the package.json or with an override, setting "main": false if this is the intention.

     Looking up github:components/jquery
ok   Installed golden-layout as npm:golden-layout@^1.0.8 (1.0.8)
     
     Bold versions are linked. To unlink use jspm install --unlink [name].
ok   Install tree has no forks.

ok   Install complete.
```

Then, when I check what was actually installed under my `jspm_packages` folder for golden-layout, this is what I see:
image

What's really odd, is if I do:
```javascript
npm install golden-layout
```

I get these directory contents in my `node_modules` folder:
image

I am **not** sure why installing via npm from jspm downloads so little when `npm install golden-layout` grabs all the necessary files.  I think it's b/c jspm can't find a *main entry point* for golden-layout when installed via jspm.

Thus, I submitted my updates above. 

I don't really want to *manage* Golden Layout's jspm registry entries, overrides, etc, so I'm open to @jedd-ahyoung, @hoxton-one, @WolframHempel correcting the situation but I would like to use golden-layout with jspm vs npm.

nkoterba:
As a follow-up, I need the css files for the UI widget in addition to the javascript.

My understanding is that by including `files` in the `package.json` or override, then **only** those files will be downloaded.  Here you can see where the golden-layout package.json specifically seems to ignore/avoid the css files:
https://github.com/hoxton-one/golden-layout/blob/master/package.json#L66-74

hoxton-one:
Hi @nkoterba ,

I'd really love to help, but I haven't used jspm yet and I know very little about it. I'm happy to accept any solution that works.

guybedford:
@nkoterba the reason golden-layout has so little is because it already has an override at https://github.com/jspm/registry/blob/master/package-overrides/npm/golden-layout%401.0.6.json.

The problem comes in this line - https://github.com/jspm/registry/blob/master/package-overrides/npm/golden-layout%401.0.6.json#L3, where the entire package is collapsed to just be the `dist` folder. Unfortunately all the other override properties are incorrect as they still use `dist`, which means they aren't referenced correctly.

Ideally we would change that override by just removing the `directories.lib` property from the override. Setting this property on overrides is discouraged these days for exactly this reason.

@2fd or @jedd-ahyoung are either of you relying on this override currently?

It would be a breaking change at this point to alter the override, although we can get away with considering it as the override is already broken.

@hoxton-one thanks for showing interest in supporting jspm - the jspm configuration in the override is exactly what needs to be put into the package.json to take ownership of this config instead of it being managed here, although it is entirely up to you if and when you want to support this.

nkoterba:
@guybedford, @hoxton-one 

Golden-Layout already has a `jspm` entry in it's package.json...it just happens to be **incorrect**.

So I see two solutions/fixes here:
1) Preferably, Golden-Layout updates it's package.json `jspm` entry to match what I submitted as the override: https://github.com/nkoterba/registry/blob/master/package-overrides/github/hoxton-one/golden-layout%401.0.8.json

This would require updating golden-layout's `package.json` and probably requiring a bump in version (1.0.8 --> 1.0.9).

We would then remove the overrides and info in the `jspm-registry` project.

@hoxton-one I can do this a pull request to golden-layout...just let me know.

2) We accept this pull request, moving golden-layout's override from the *broken* 1.0.6 version currently in place to the 1.0.8 version I submitted.

hoxton-one:
Hi @nkoterba, thank you very much for your effort, much appreciated. I think a pull request to GoldenLayout and a subsequent new version release might be less intrusive, shall we go with that?

nkoterba:
@hoxton-one I think that's the most "ideal" situation. I will work a pull request for GL now...

guybedford:
This sounds like a good approach. As soon as that lands, we can clear out the override here to be an empty JSON file so that the package one takes preference.

PR #370, opened 2015-05-03
Add headroom.js registry and override

smajl:


guybedford:
Note this should use a `.json` extension.

Also we no longer use `directories` properties in overrides for backwards-compatibility.

smajl:
I fixed the extension and added global export to be safe.

But, what about `directories`? Should I use something like:
```json
"main": "dist/headroom",
"files": ["dist/**/*"]
```
or what? I don't get it. Can't find anything new in docs about that.

guybedford:
@smajl yes exactly we can leave out directories and set the main to `dist/headroom`.

smajl:
I guess it should be OK now.

PR #371, opened 2015-05-05
Adding Dimple (D3-based charting library)

geelen:
Dimple is hosted on Github, but doesn't have a `dimple.js` or `index.js` in the project. But we can directly reference the latest version without too much trouble (or should we use `dimple.latest.min` instead?)

### DIFF
--- a/registry.json
+++ b/registry.json
@@ -130,6 +130,7 @@
   "decimal.js": "github:MikeMcl/decimal.js",
   "dgram": "github:jspm/nodelibs-dgram",
   "di": "github:angular/di.js",
+  "dimple": "github:PMSI-AlignAlytics/dimple/dist/dimple.latest"
   "dns": "github:jspm/nodelibs-dns",
   "domain": "github:jspm/nodelibs-domain",
   "dropzone": "github:enyo/dropzone",


registry.json
```diff
@@ -130,6 +130,7 @@
   "decimal.js": "github:MikeMcl/decimal.js",
   "dgram": "github:jspm/nodelibs-dgram",
   "di": "github:angular/di.js",
+  "dimple": "github:PMSI-AlignAlytics/dimple/dist/dimple.latest"
```
guybedford: Trailing comma!

guybedford:
Merged manually in https://github.com/jspm/registry/commit/67c8cdcb3c7a760907f99cd2ff7453d0669e5bb8. The main should be picked up in the bower.json I believe, but if it isn't working, we can add an override to set that.

geelen:
Thanks! Still getting a 404 on https://registry.jspm.io/dimple.js, should I be waiting for a deploy?

guybedford:
Sure I've updated the cache on the registry now.

geelen:
Ok, so it's not picking it up from bower, now getting a 404 on https://github.jspm.io/PMSI-AlignAlytics/dimple.js when I `import dimple from 'dimple'`

`import dimple from 'dimple/dist/dimple.latest'` works though

guybedford:
Yeah that's actually a bug. I've added an override in https://github.com/jspm/registry/commit/342f431809ad72b5a546174b12cabeacb2e16788 to set this, will aim to get a fix in though.

PR #377, opened 2015-05-08
initial version of TypeScript 

vladima:
I'll send a separate PR to update TypeScript registry entries after the official release of 1.5


guybedford:
This is great, thanks. Is it worth including this configuration in TypeScript's package.json itself? Happy to keep it here otherwise though.

PR #380, opened 2015-05-10
Angular material 0.9.0 overrides added

red010b37:
This purposely does not define angular dependencies to allow the users to specify project angular dependencies versions (e.g 1.3.15 or 1.4.0 ) - This follows the Angular teams last override which removed all angular dependencies.

This commit however does add the css shim for a better usage pattern inside JSPM

### DIFF
--- a/package-overrides/github/angular/bower-material@0.9.0.json
+++ b/package-overrides/github/angular/bower-material@0.9.0.json
@@ -1,0 +1,12 @@
+{
+	"main": "index",
+	"registry": "jspm",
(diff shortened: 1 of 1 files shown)


guybedford:
The override here is managed in the bower repo at https://github.com/angular/bower-material/blob/master/package.json. I'd suggest having the discussion there.

red010b37:
Ok thanks for that - I will go and have a chat

red010b37:
hwo

PR #384, opened 2015-05-14
clarify a couple of settings

codylindley:


guybedford:
Looks great - while we're about this can we make the main the unminified source? It can be a better option for debugging.

codylindley:
Sure. Done.

PR #385, opened 2015-05-15
Add Angular2 to registry

robwormald:
Adds overrides for:
- ES6 format
- using es6 source files
- TODO: have ng2 output .js instead of .es6 extensions here.


johnjelinek:
Should you include the angular 2 router with this?

robwormald:
@johnjelinek at the moment the router lives in the main angular2 codebase: https://github.com/angular/angular/blob/master/modules/angular2/router.js  - so this PR would allow you to import from angular2/router.

Once these features break (back) into their own repos, it would make sense to have distinct entries in the registry.

johnjelinek:
:+1:



—

djindjic:
:+1:

guybedford:
@robwormald nice! Is this good to merge, or are the `.es6` extensions a show-stopper still? Working hard on getting the SystemJS release out that will allow more flexibility here.

robwormald:
@guybedford I reckon we should wait until the next rev, and get those extensions sorted. I couldn't find a good way to add them to the registry and have them propagate when installed. 

I'm working on a PR angular side to adjust.

guybedford:
@robianmcd sure, do keep me posted on how it is going there, and ping me if you need any feedback at all.

robianmcd:
@guybedford I think you meant to mention @robwormald . But by some crazy coincidence I just started trying to setup a seed project using jspm and Angular2 today!

guybedford:
Haha, sorry about that, glad to hear it wasn't too inappropriate!

robwormald:
@guybedford cool, thanks. since its related, whats your plan on TypeScript support? is it going to be a first-class transpiler option ( eg System.transpiler = 'typescript' )? 

It would also be useful to maybe have an "extension" config option for packages - doing .es6 seems like a bad choice, but there's a number of libraries out there that are doing this, and of course you'd need the same kind of functionality to enable seamless .ts import / compilation.

andreasgrimm:
support for TypeScript would be awesome.

guybedford:
Yes TypeScript will have first-class support thanks to the work by @vladima in https://github.com/jspm/jspm-cli/pull/735. We're working on merging this in today, to be released along with this next major.

djindjic:
nice news! thanks @vladima and @guybedford

crissdev:
I hope `.es6` suffix won't survive long. If people want to distinguish ES6 from ES5 they can easily name their files like `file.es6.js`. Or maybe event better `file.es7.js` :smile:

gdi2290:
:+1:

jimschubert:
:+1:

robwormald:
@guybedford and all - now would seem an appropriate time to land this properly, with the changes in the new version... any thoughts on how to best handle it?

guybedford:
@robwormald unfortunately despite the feature to load any extension in SystemJS, the defaultJSExtensions compatibility mode does hold jspm back for a little while longer until the upgrade path to remove that. Was going to do it in one, but felt it better to get it out the door instead of continuing to hold users back.

Shall I still go ahead and land this, or was the es6 extension support the hope?

robwormald:
we can probably get the extension fixed, i guess its more a question of what the "right" way to support a typescript source module when its going to be consumed through jspm (and really what users are going to prefer to consume). 

Use JSPM to consume .ts source and build from there? Publish System.register bundles? Multiple registry entries (angular2 and angular2-ts?)? Start another registry called tspm? ;-)

guybedford:
@robwormald separate file System.register compilation is probably the most advisable. But it does introduce publishing overhead. tspm works for me too :)

guybedford:
Note by separate file, I mean compiling each file individually into an anonymous System.register. Folding can be suitable though too, depending on the the granularity needed for consumers. For example do you want consumers to be able to load partial parts of angular (`import 'angular/feature'`) or is it always a single module object?

robwormald:
Okay cool - I think it makes sense to do the anonymous thing. Presumably they can then be (re) bundled by consumers of the module with jspm / systemjs builder?

guybedford:
If using TypeScript there are runtime concerns for this type of output. The reason being that each small module file will inline its helpers. Babel provides a modular runtime so we can load these runtime pieces as modules themselves in a way that respects bundling well, but even then one has versioning concerns for the runtime itself as it updates over time.

Sorry to go back on what I said but with that in view perhaps distributing the original source or a folded bundle may be the better bet here. In the first case users get to share the helpers with their final bundle. In the second case, helpers are at least shared within the entire codebase of Angular instead of in each file.

Let me know if that makes sense, still very much feeling out the best practises.

robwormald:
https://github.com/angular/angular/pull/2455 nearly there!

robwormald:
alpha 27 is out, with .js extensions for all outputs (other than ts of course), so i've set this back to the defaults - the override is just dependencies now. 

the npm distribution has
- commonJS in the root directory (this will be used by default)
- es6 source in es6/dev and es6/prod - the dev version does extra debug stuff in lieu of performance.
- typescript source in ts/ with .ts extensions.
- typescript definitions in all directories

@guybedford this is ready to go, and should let people experiment a lot easier. its also a pretty good test case for package configuration scenarios, like toggling between dev and prod mode or using the typescript source.

guybedford:
@robwormald excellent, it's great to have this in as a simple install!

PR #386, opened 2015-05-15
Add Immutable and ImmutableJS mappings

jackfranklin:
@guybedford what do you think? Kind of makes sense IMO given that React is in the registry and a lot of people work on them together.


guybedford:
Sure, always happy to add things here. Perhaps let's leave out the `immutablejs` option or is that the proper name? Nice to have one option to keep it simple.

jackfranklin:
:+1: Wasn't sure on this - it's on npm as `immutable` so let's just add that.

guybedford:
Great, thanks.

PR #396, opened 2015-05-28
Update angular-formly overrides and add api-check dependency

kentcdodds:
Didn't notice this override. I'm not certain this is how things work, but basically angular-formly has both of these dependencies listed as peerDependencies and I'm guessing that https://github.com/formly-js/angular-formly/issues/321 is because jspm doesn't resolve peerDependencies correctly.

### DIFF
--- a/package-overrides/github/formly-js/angular-formly@5.0.3.json
+++ b/package-overrides/github/formly-js/angular-formly@5.0.3.json
@@ -1,5 +1,0 @@
-{
-  "dependencies": {
-    "api-check": "github:kentcdodds/apiCheck.js@^6.0.11"
-  }
-}
--- a/package-overrides/npm/angular-formly@6.10.1.json
+++ b/package-overrides/npm/angular-formly@6.10.1.json
@@ -1,0 +1,6 @@
+{
+  "dependencies": {
+    "api-check": "^7.0.0",
+    "angular": "^1.2.x || >= 1.4.0-beta.0 || >= 1.5.0-beta.0"
+  }
+}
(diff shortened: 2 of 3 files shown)


guybedford:
Ok it sounds like this issue is due to jspm not supporting arbitrary range operators which is actually resulting in a fork of Angular.

I've reverted the npm switch for now, but we can run on either actually.

Note that instead of including an override you can simply add jspm configuration to the package.json which is preferable when you have access. The override here can then be blanked out entirely.

There's an issue on jspm to support peer dependencies, which will stop these kinds of conflicts from happening - that will be the ultimate solution to this issue.

For now I'd suggest leaving things as they are, or giving jspm a more flexible angular range in the package.json:

```json
{
  "jspm": {
    "peerDependencies": {
      "angular": "*"
    }
  }
}
```

kentcdodds:
Hmmm... so, angular-formly has two dependencies: `angular` and `api-check`. Currently in the `package.json` it has a `peerDependencies` with quite loose semver requirements on each.

If I understand correctly, you're saying that jspm doesn't support the built-in `package.json` `peerDependencies` property, but if I specify it under `jspm` with a wildcard, that will fix the issue? So, it should look like this (for now):

```json
{
  "jspm": {
    "peerDependencies": {
      "angular": "*",
      "api-check": "*"
    }
  }
}
```

Right?

guybedford:
@kentcdodds the issue is that jspm doesn't support `||` in version ranges currently, resulting in a fork of Angular.

jspm will take peerDependencies and turn them into normal dependencies fine, so we just need to override the range specifically.

What you have there should work though too (the api-check as a peer dependency will just override the normal api-check dependency I believe).

kentcdodds:
Should be working now.

PR #398, opened 2015-05-30
Create angular-hotkeys@1.4.5.json

lookfirst:


package-overrides/github/chieffancypants/angular-hotkeys@1.4.5.json
```diff
@@ -0,0 +1,15 @@
+{
+  "directories": {
+    "lib": "build"
```
guybedford: We're not encouraging this anymore for exactly the reason that it can break things for others as you experienced.

guybedford:
Thanks, just let me know when you've got the correction in.

lookfirst:
Yea, I had just copy / pasted the other one in the same directory. You might want to update that one too.

package-overrides/github/chieffancypants/angular-hotkeys@1.4.5.json
```diff
@@ -0,0 +1,12 @@
+{
+  "main": "hotkeys",
```
guybedford: This should be `src/hotkeys.js` or `build/hotkeys.js`

package-overrides/github/chieffancypants/angular-hotkeys@1.4.5.json
```diff
@@ -0,0 +1,12 @@
...
+  "registry": "jspm",
+  "dependencies": {
+    "angular": "^1.2.0"
+  },
+  "shim": {
+    "hotkeys": {
```
guybedford: This should be `src/hotkeys` or `build/hotkeys`.

guybedford:
Ah right, best to leave the other one now that it is in use though.

lookfirst:
On a side note, it would be nice if any code that has the word 'angular.', in it would automatically depend on angular. ;-) I know that is a magic rathole you probably don't want to go down, but shrug... it's friggen javascript at the end of the day...

guybedford:
Automatic detection of globals is a good route for automatic override creation workflows. It's just a matter of putting it together.

PR #399, opened 2015-05-31
Adds auth0/lock override.

Zoltu:
Copied from https://github.com/volak/registry/blob/master/package-overrides/github/auth0/lock%407.1.1.json, seems to work.


guybedford:
Thanks - note the override must end in `.json`, and be valid a json file.

Zoltu:
Fixed I believe.

PR #402, opened 2015-06-02
added zepto

Aaike:


guybedford:
Did you try installing from npm? What is wrong with https://github.com/madrobby/zepto? Are you sure you want to maintain this indefinitely?

Aaike:
the github didn't have the distribution files.
And i totally forgot about the npm version ...
that is exactly what i needed... thx !

PR #408, opened 2015-06-05
mongoose added to registry

capaj:
@gavinaiken after this is merged, you should be able to install mongoose with just:
```
jspm i mongoose
```


guybedford:
Thanks @capaj!

capaj:
although it doesn't work with jspm 0.15.7. Only with 0.16.0. Seems like I will be staying on beta from now...

PR #412, opened 2015-06-07
Adds support for STRML/react-router-component.

brendoncrawford:
There is one small issue, which is given here:

``` javascript
import * as ReactRouter from `react-router-component`;
```

For some reeason, the actual module is
located at `ReactRouter.default`. Is this a system.js issue?

### DIFF
--- a/package-overrides/github/STRML/react-router-component@0.25.1.json
+++ b/package-overrides/github/STRML/react-router-component@0.25.1.json
@@ -1,0 +1,8 @@
+{
+  "main": "index",
+  "registry": "npm",
+  "format": "cjs",
+  "directories": {
+    "lib": "./"
+  }
+}


guybedford:
@brendoncrawford thanks, should we not be using the `npm:react-router-component` version though?

Note CommonJS imports are via `import ReactRouter from 'react-router-component'`. Named exports will work as well in the latest jspm beta though.

brendoncrawford:
@guybedford : I did not use the npm version, because it does not include the `index.js` file, which I thought was a convenient entry point for the module.

guybedford:
@brendoncrawford can you clarify what you mean that it does not include the index.js file? How is it supposed to work on npm otherwise?

guybedford:
I just tried installing it using npm and there is definitely an index.js file for verstion 0.25.1, is this a jspm support issue for this perhaps?

brendoncrawford:
@guybedford: Ok, my original statement was incorrect.  Here is what I meant to say:

When I use npm instead of github, then when I import it, the module which gets imported is `./lib/Router.js` instead of `./index.js`.  I have no idea why this is the case.

guybedford:
@brendoncrawford `jspm install npm:react-router-component` is working fine for me actually. Have you tried it on the latest jspm beta?

brendoncrawford:
@guybedford: Ok, when I upgraded to jspm beta, using the npm registry seems to work fine. I will update this pull request sometime in the next 1 - 2 days accordingly. Thanks for helping with this.

PR #413, opened 2015-06-07
fixed Dogfalo/materialize package override for case-sensitive systems

manuel-guilbault:
renamed package override ```Dogfalo/materialize``` to ```dogfalo/materialize```, as it caused issues on case-sensitive systems, because the package is named ```dogfalo/materialize``` everywhere else.


guybedford:
Can you clarify what you mean by it being named `dogfalo/materialize` everywhere else?

manuel-guilbault:
Well in registry.json, the path is lower-cased: ```github:dogfalo/materialize```.

manuel-guilbault:
The issue is that when installing "materialize", it uses "dogfalo/materialize" as the package name, and on a case -sensitive system like Ubuntu, the package override is not found because it is cached as "Dogfalo/materialize"

guybedford:
I've added https://github.com/jspm/registry/commit/d2dd4a0ca95da33b615d556ce8e343cbcb4237cf. Let me know if that change helps as an alternative?

manuel-guilbault:
Yes, it fixes the issue. Thanks!

PR #414, opened 2015-06-09
Tinyanimate

Aaike:


package-overrides/github/branneman/TinyAnimate@0.2.0.json
```diff
@@ -0,0 +1,6 @@
+{
+  "main": "TinyAnimate.js",
+  "directories":{
+    "lib": "bin"
```
guybedford: We're discouraging this for backwards-compatibility. I know the chances are small an existing user is relying on this already, but it's not a chance we can take.

PR #419, opened 2015-06-12
Added PureMVC Standard (v1.0a) and Multicore (v1.1) repositories.

sparkbuzz:


registry.json
```diff
@@ -304,7 +306,7 @@
   "theseus": "npm:theseus",
   "theseus-angular": "npm:theseus-angular",
   "topo": "github:capaj/topo-browser",
-  "timers": "github:jspm/nodelibs-timers",
+  "timers": "github:jspm/nodelibs-timers",c
```
guybedford: Can you remove this character change? I think that may be invalidating the JSON.

sparkbuzz: Thanks for pointing out @guybedford , I've removed the char, must have entered it by mistake.

registry.json
```diff
@@ -120,7 +120,7 @@
   "chai": "npm:chai",
   "chart": "github:nnnick/Chart.js",
   "child_process": "github:jspm/nodelibs-child_process",
-  "chosen": "github:harvesthq/chosen",
+  "chosen": "github:harvesthq/chosen",p
```
guybedford: There's another character here as well!

guybedford:
Thanks, looks like there's one more though.

sparkbuzz:
Sorted my issues out, once merged, I'll retest to see if PureMVC installs correctly.

PR #421, opened 2015-06-14
Added github:Upstatement/jquery-total-storage

mtiitson:


guybedford:
Note the file format should be json instead of js.

PR #425, opened 2015-06-15
Investigating failure to find entry point.

sparkbuzz:
I'm really not sure why `jspm install` cannot find the entry point, some advice would be highly appreciated.


guybedford:
See https://github.com/jspm/registry/commit/93a5f4dcfe011df64100906f8e5a6c4da00325bc.

sparkbuzz:
@guybedford Thanks for the change you've made.

I see you didn't merge the change I made on this PR, I see all the other libs are omitting .js so I removed mine here too. Shouldn't the .js be omitted rather, for the sake of consistency?

guybedford:
The js extension is optional, but it is better to include it.

PR #434, opened 2015-06-19
Add svg.js registry and override

MeirionHughes:


guybedford:
Great, thanks. Note for future reference usually we try to reference unminified files as it allows for a better debugging experience, then suitable minification can be done at the end.

PR #442, opened 2015-06-25
add override to moment to make it global

lookfirst:


guybedford:
This is a common issue - everything needing to become a global to be compatible with globals.

We fixed this in the new globals config in SystemJS so this won't be necessary in future. But support for `globals` in the package.json will come with the jspm@0.17 configuration update to match the new SystemJS features.

lookfirst:
@guybedford Thanks! I've got another PR coming then.... =)

PR #451, opened 2015-07-02
Create HumanizeDuration@3.1.0.json

lookfirst:
not sure why i have to specify a version again...


guybedford:
You shouldn't need to post a new version. I think the issue is that the repo has a ".js" at the end of the name, so we should probably rename the version from https://github.com/jspm/registry/tree/master/package-overrides/github/EvanHahn.

PR #453, opened 2015-07-03
add js-cookie

Aaike:


guybedford:
I don't know how this slipped through in my notifications - apologies for the delay!

PR #456, opened 2015-07-05
sass.js: Ignore 'ws' dependency brought in by emscripten but never used

theefer:
See https://github.com/medialize/sass.js/issues/34


theefer:
Should hopefully fix https://github.com/theefer/plugin-sass/issues/2

PR #457, opened 2015-07-05
add shim for the ui pakage

Aaike:


package-overrides/github/julianshapiro/velocity@1.1.2.json
```diff
@@ -5,5 +5,12 @@
     "velocity.min.js",
     "velocity.ui.js",
     "velocity.ui.min.js"
-  ]
+  ],
+  "shim": {
+    "velocity.ui": {
+      "deps": [
+        "github:julianshapiro/velocity"
```
guybedford: It's probably best to just reference the main entry point explicitly here.

Note also this request as you've written it wouldn't work against jspm 0.16 beta.

Aaike: oh ok , i guess i need to read up on the beta notes :)

but the `velocity.ui` is not the main file.
velocity works just fine as it is with the current override , but the `velocity.ui` is an extra file that can optionally be loaded in, it seems to need the main file as a dependency.

PR #472, opened 2015-07-14
fix dep for css file

lookfirst:
that said, it doesn't seem to really do anything. i don't see how this works. @guybedford please take a look?

### DIFF
--- a/package-overrides/github/chieffancypants/angular-loading-bar@0.8.0.json
+++ b/package-overrides/github/chieffancypants/angular-loading-bar@0.8.0.json
@@ -6,8 +6,8 @@
     "css": "*"
   },
   "shim": {
-    "loading-bar": {
-      "deps": ["angular", "build/loading-bar.css!"]
+    "build/loading-bar": {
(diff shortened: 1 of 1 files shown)


package-overrides/github/chieffancypants/angular-loading-bar@0.8.0.json
```diff
@@ -7,7 +7,7 @@
   },
   "shim": {
     "loading-bar": {
```
guybedford: `build/loading-bar`

package-overrides/github/chieffancypants/angular-loading-bar@0.8.0.json
```diff
@@ -7,7 +7,7 @@
...
-      "deps": ["angular", "./loading-bar.css!"]
+      "deps": ["angular", "build/loading-bar.css!"]
```
guybedford: `./loading-bar.css!` is right here.

lookfirst:
@guybedford ping

PR #475, opened 2015-07-16
Added TinyMCE and overrides

Munter:
I am in doubt on ho this works.

How do I get my override to match more semvers than just the one that I had to hard code into the file name?

Do I need to set anything else in the package override?

### DIFF
--- a/package-overrides/github/tinymce/tinymce-dist@4.1.7.json
+++ b/package-overrides/github/tinymce/tinymce-dist@4.1.7.json
@@ -1,0 +1,12 @@
+{
+  "format": "global",
+  "main": "tinymce.min",
+  "shim": {
+    "tinymce.min": {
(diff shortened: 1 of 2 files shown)


guybedford:
Looks good! Usually try to reference unminified for development experience, over using minified files in the package configuration. Idea is that production-level optimization would do the minification. Overrides apply for all future semver-compatible versions, it's mentioned at https://github.com/jspm/registry#submitting-a-package-override.

Munter:
Future compatible given what semver range query? `4.1.7` is not compatible with any other version. `4.x` would be compatible with all minor and patch versions under major version 4.

Should the file names stop using patch versions explicitly and go directly for a wider semver range?

guybedford:
It will apply to all `^4.1.7`.

PR #477, opened 2015-07-16
Tinymce

Munter:
Based on better understanding of file naming I'm moving this override to cover versions 4.x.x

Also added shims for all plugins


package-overrides/github/tinymce/tinymce-dist@4.0.0.json
```diff
@@ -0,0 +1,258 @@
...
+  "main": "tinymce",
+  "shim": {
+    "tinymce.min": {
+      "exports": "tinymce"
+    },
+    "tinymce": {
+      "exports": "tinymce"
+    },
+    "plugins/advlist/plugin": {
+      "imports": "tinymce",
```
guybedford: `imports` is not a valid shim property. I think we might have used `imports` once though in the past and it may still be supported for backwards compat, but best to just use `deps`.

package-overrides/github/tinymce/tinymce-dist@4.0.0.json
```diff
@@ -0,0 +1,258 @@
...
+      "deps": [
+        "tinymce"
```
guybedford: Does this definitely work? It may be necessary to do backtracking here - `../../tinymce`. Deps are written as if it was a require from within the module file.

Munter: It works. But may break if the user saves the package under a different label?

package-overrides/github/tinymce/tinymce-dist@4.0.0.json
Munter: Saw it in another override. Fixed now

PR #480, opened 2015-07-17
Add toastr w/ override

PWKad:


PWKad:
Crud one second this needs to depend on the css as well give me a minute...

PWKad:
@guybedford I've tried this locally but it doesn't pull down the CSS by default - 

```
jspm install github:CodeSeven/toastr  -o "{ shim: { 'build/toastr.min': { deps: ['jquery', './toastr.min.css\!'] } }, dependencies: { jquery: '*', 'css': '*' } }"
```

Should it?  Sorry for the stupid question, I'll fix this PR once I get it right >.<

guybedford:
Yes that should work. Is `build/toaster.min.js` definitely the main?

PWKad:
Yeah the js is working fine just can't seem to get it to pull down the CSS without a separate import.  Maybe that's normal (I've not written these package overrides much before)

guybedford:
The way to check is to see if the file at `jspm_packages/github/CodeSeven/toastr/build/toastr.min.js` contains `"deps ./toastr.css!"` at the top of it.

PWKad:
Interesting it sure does but still need to use  - 

```
import toastr from 'CodeSeven/toastr';
import 'CodeSeven/toastr/build/toastr.min.css!';
```

But at least it's working that way.

guybedford:
The only reason it wouldn't work with that then was if it was not being interpreted as a global module.

guybedford:
I just had a look at this and the file being used is `toastr` not `build/toastr`. Changing the shim to `toastr` instead should make this work.

PWKad:
I'm cleaning up old PR's should I close this one?  Looks like someone else has a merged PR already for toastr.

guybedford:
Sure, thanks for following up.

Issue #483, opened 2015-07-18
what is the override config for https://github.com/weluse/hyperagent ?


docteurklein:

Munter:
It's probably going to be hard to deal with that `Hyperagent.configure(...)` thing. But it should be possible to override in a way so the assumed defaults `window.$` and `window.Q` are available. I'll take a look at it

docteurklein:
Thanks a lot for your response :)

I guess I can live with a manual invokation of configure if necessary: In fact configure is just optional.

My biggest concern was how to configurethe shims and co.
I tried hard to modify my jspm config inside packages.json, but it looked like the generated config.js didn't care at all about my overrides.
Anyway, thanks :)

Munter:
Took me a while to figure out how overrides even affect things. I think this need a more explicit workflow an cause/effect explanation. Overrides actually mostly add comment annotations in the downloaded source code, which can then instruct the loader to do the right thing. This is cool because it doesn't clutter up your local config, but on the other hand without a clear explanation of what it does it feels a bit like black magic.

Munter:
So this gives me the correct dependencies in that it also installs jquery and q when installing, it loads jquery and q when loading the package. But I can't figure out the missing config to force jquery and q to expose globals, which will then be picked up by hyperagent. It picks up jquery because jquery apparently always leaks itself to `$` in window (bad jquery), but Q doesn't do that.

```
jspm install hyperagent=github:weluse/hyperagent -o "{ main: 'dist/hyperagent.js', format: 'global', dependencies: { jquery: '*', q: 'npm:q' }, shim: { 'dist/hyperagent': { deps: ['jquery', 'q'] } } }"
```

docteurklein:
wow you rock!

Maybe we should override Q, with a shim and a global format ?

docteurklein:
BTW @Munter I don't know if it helps, but both BAMD and cjs are present in the dist folder.

Munter:
That would be bad practice imo. Configuring another package to default to doing something different. It would probably be best if the needed configuration was encapsulated in hyperagent itself. The behavior we are looking for is something along these lines:

```js
System.import('jquery')
  .then($ => { window.$ = $})
  .then(() => { return System.import('q') })
  .then(q => { window.Q = q })
  .then(() => [ return System.import('hyperagent') })
  .then(hyperagent => {
    // Finally!
  });
```

Munter:
[...]

PR #486, opened 2015-07-19
Added hyperagent

Munter:
There is still an outstanding question in #483 on how to get an override to leak a global from a dependency.

### DIFF
--- a/package-overrides/github/weluse/hyperagent@0.1.0.json
+++ b/package-overrides/github/weluse/hyperagent@0.1.0.json
@@ -3,13 +3,15 @@
   "format": "global",
   "dependencies": {
     "jquery": "*",
-    "q": "npm:q"
+    "q": "npm:q",
+    "uri": "github:medialize/URI.js"
   },
   "shim": {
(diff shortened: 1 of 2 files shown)


guybedford:
Thanks! Do you mean supporting the `globals` meta as part of the override here? That will be in 0.17, but 0.17 is a really hard release. Working on it!

guybedford:
Perhaps we could support globals in the mean time within the existing configuration system as `shim.globals` in the package config (statically writing into `"globals.jQuery jquery"` etc. It should actually be fairly trivial to implement - if you're interested in a PR along these lines, would be happy to point to the lines of code to implement it.

Munter:
Yeah, that sounds like a good addition. I'll be happy to take a look with some guidance

guybedford:
Awesome. The shim meta-injection is handled in https://github.com/jspm/jspm-cli/blob/master/lib/build.js#L421. It should be straightforward to add a `globals.jQuery: 'jquery'` meta to this shim object, which would then work with SystemJS 0.18.

PR #490, opened 2015-07-20
switch to use TypeScript npm package (1.5.3)

vladima:


guybedford:
Thanks so much for getting to these so promptly! Let's keep the old one in place for existing users here, and add a new one rather.

guybedford:
`jspm install npm:typescript` actually seems to work absolutely fine. It may be installing unnecessary Node dependencies, which could be worth investigating for an override. Also it could be worth checking if the Browserify version or the global version is better to run. I did a quick test and the direct `npm:typescript` Node / Browserify version works fine in the browser and Node for me.

guybedford:
Added https://github.com/jspm/registry/commit/f5b907355bfb4975195fff4c69ea19a9942713ae.

vladima:
@guybedford done, I've kept the old file and added a new set of overrides for npm version

guybedford:
@vladima any thoughts on whether to use the npm version plain or with the override? Both seem to work I guess.

My thinking is it may be better to use the npm version plain without any override as that means we don't need to maintain the override, and also that TypeScript then works in jspm both client and server (while the override makes it only work in the browser not in Node I believe).

Let me know what you think?

vladima:
@guybedford Ideally I'd like things to just work without overrides. However there is an [issue on the TypeScript side] (https://github.com/Microsoft/TypeScript/issues/3408) and I think we might need overrides to deal with it. At this point I don't think we can just move package entry point to another folder since it will be a breaking change for anyone who references the file via relative path. Overrides looked like an acceptable solution for JSPM case

guybedford:
@vladima I'm wondering if we should just add the `web.config` file by default for jspm. The reasoning being that anything published to a registry is by definition not sensitive. So that we do want everything servable. Handling this in jspm may make sense if it smooths this use case.

The main reason I worry about this is that in jspm, we use the same version of Babel and Traceur to load in the client as we do to bundle (not the `node_modules` version, that `jspm_packages` version). So if we want the simple TypeScript bundling workflow to work in jspm (without needing an npm install exception manually coded in), using the npm version of TypeScript as it stands seems to make the most sense to me.

Interested to hear your opinion further on this.

vladima:
the issue with hidden segments is they are coarse grained: removing `bin` from the set will allow access to all folders named `bin` on the website so both `typescript/bin/typescript.js` and <webroot>/bin/SomeDllWithSensitiveData.dll will be accessible. A solution for this is quite clumsy and boils down to: remove `bin` from hidden segments then block access to every individual `bin` folder - I would prefer not to go this way.

We can augment TypeScript npm package in the following way and by creating folder named `dist`, copy the content of `bin` into `dist` and use overrides to specify `main` as `dist/typescript.js`. This way we can preserve backward compatibility for anyone who used relative name with `/bin/` in it to reference typescript.js. The cost of this solution is doubling the size of the package (from ~11 MB to ~22 MB) which is quite a lot. As an option we can copy not the `bin` wholesale but just a few files (ideally one which is `typescript.js`) but in this case it will make the life more complex for people who obtain TypeScript via jspm to do normal compilation\language service related work - standard library with typings will no longer be next to `typescript.js`.

vladima:
as an option we can still ship everything in `bin` but add a `postinstall` script that will create `dist` folder and fill it with symlinks for all files in `bin`.

vladima:
@guybedford added a fix proposal [here](https://github.com/Microsoft/TypeScript/issues/3408#issuecomment-124194402)

guybedford:
@vladima ok great, so shall we go ahead with using the npm:typescript in the mean time then? The `bin` issue will only affect a minority of users I expect, and there is a workaround anyway via that configuration option.

vladima:
@guybedford yes, that should be fine. We've discussed this issue offline and decided to move a .js content of npm package to lib (https://github.com/Microsoft/TypeScript/pull/4039). Once this change goes in I'll send PR with matching changes to es6-module-loader/systemJS

guybedford:
Ok great. Closing this PR then as I believe our `npm:typescript` is working ok pending that change. But if there is anything else we need to look at for it please let me know.

PR #491, opened 2015-07-22
Create moment-timezone@0.4.0.json

lookfirst:


Munter:
If you use `npm:moment-timezone` this will already work. Maybe we should register that endpoint in the registry instead?

guybedford:
@lookfirst let me know if you've tested out the npm version.

lookfirst:
I try to never use the npm version of things as I like the option to fork and because there is no cross repo dep resolution. So, please just merge this.

Jon

lookfirst:
just to give you and idea of how github focused i am...

my current package.json.... 

```
    "dependencies": {
      "accounting": "github:openexchangerates/accounting.js@^0.4.1",
      "angular": "github:angular/bower-angular@^1.4.3",
      "angular-animate": "github:angular/bower-angular-animate@^1.4.3",
      "angular-carousel": "github:revolunet/angular-carousel@^0.3.12",
      "angular-datetime": "github:eight04/angular-datetime@^1.2.3",
      "angular-deferred-bootstrap": "github:philippd/bower-angular-deferred-bootstrap@^0.1.7",
      "angular-input-modified": "github:betsol/angular-input-modified@^2.2.3",
      "angular-loading-bar": "github:chieffancypants/angular-loading-bar@^0.8.0",
      "angular-messages": "github:angular/bower-angular-messages@^1.4.3",
      "angular-mocks": "github:angular/bower-angular-mocks@^1.4.3",
      "angular-motion": "github:mgcrea/angular-motion@^0.4.2",
      "angular-redactor": "github:TylerGarlick/angular-redactor@^1.1.5",
      "angular-resource": "github:angular/bower-angular-resource@^1.4.3",
      "angular-sanitize": "github:angular/bower-angular-sanitize@^1.4.3",
      "angular-socialshare": "github:720kb/angular-socialshare@^0.1.9",
      "angular-strap": "github:mgcrea/angular-strap@^2.3.0",
      "angular-timer": "github:siddii/angular-timer@^1.3.3",
      "angular-touch": "github:angular/bower-angular-touch@^1.4.3",
      "angular-ui-router": "github:angular-ui/ui-router@^0.2.15",
      "babel-core": "npm:babel-core@^5.7.2",
      "clean-css": "npm:clean-css@^3.3.4",
      "css": "github:systemjs/plugin-css@^0.1.6",
      "hotkeys": "github:chieffancypants/angular-hotkeys@^1.4.5",
      "interact": "github:taye/interact.js@^1.2.4",
      "jquery": "github:components/jquery@^2.1.4",
      "jsog": "github:jsog/jsog@^1.0.7",
      "json": "github:systemjs/plugin-json@^0.1.0",
      "lodash": "npm:lodash@^3.10.0",
      "moment": "github:moment/moment@^2.10.3",
      "moment-timezone": "github:moment/moment-timezone@^0.4.0",
      "ng-file-upload": "github:danialfarid/ng-file-upload@^5.0.9",
      "ngTagsInput": "github:mbenford/ngTagsInput@^3.0.0",
      "phoneformat": "github:albeebe/phoneformat.js@^1.0.1",
      "selectize": "github:brianreavis/selectize.js@^0.12.1",
      "ui-router-extras": "github:christopherthielen/ui-router-extras@^0.0.14",
      "ui-select": "github:angular-ui/ui-select@^0.12.0"
    },
```

lookfirst:
@guybedford Actually, this is quite interesting. I just tried to import moment-timezone and it failed because index.js looks like this:

```
/* */ 
"format cjs";
var moment = module.exports = require("./moment-timezone");
moment.tz.load(require('./data/packed/latest.json'));
```

What ends up happening is that a .js extension is appended to the latest.json and everything blows up.

I'm on latest jspm beta btw.

guybedford:
Interesting, I'm definitely getting  

```
/* */ 
var moment = module.exports = require("./moment-timezone");
moment.tz.load(require("./data/packed/latest.json!systemjs-json"));
```

myself, being converted by the node conversion layer in the npm registry implementation.

Are you sure you're running the latest jspm-npm? It may also be worth trying a force install in case this is a stale cache issue.

lookfirst:
I'm getting it with the github endpoint not npm.

guybedford:
@lookfirst right, yes requiring json files is only supported via the npm registry conversion process.

lookfirst:
What is the thing preventing it from working with github? Let's fix that...

guybedford:
Requiring a json file without any further config is not supported in jspm. It's an npm convention, so makes no sense to treat by default on GitHub.

lookfirst:
What I hear you saying is that jspm only works with github repositories on monday, wednesday and friday. Forget tuesday and thursday.

That makes absolutely no sense.

guybedford:
@lookfirst `require('file.json')` is not supported in jspm full stop. If you want to support it write a Node module and use the Node ecosystem.

lookfirst:
@guybedford I'm looking at this file: https://github.com/moment/moment-timezone/blob/develop/index.js

It is clearly being rewritten by jspm, no? Why not rewrite it with the same thing that you use to rewrite it via jspm-npm? What is the harm?

guybedford:
Because it isn't just that one convention. The following things happen in Node requires:

* `require('./dir')` will load `dir/index.js` or `dir/index.json`
* `require('./file')` will load `file.js`, `file.json`, `file` in that order

etc etc

There are a bunch of these conventions which are specific to the Node resolution algorithm (https://nodejs.org/api/modules.html#modules_all_together), and supporting them is part of a Node compatibility layer in the npm registry.

We can't then assume that all CommonJS modules on the internet use this exact same resolution algorithm. And loading from GitHub means removing that Node assumption.

If you really want to install something from GitHub with the Node conversion layer, you can simply do `jspm install github:package/name -o "{registry: 'npm'}"`. But it is then better to install the `npm:` equivalent anyway.

lookfirst:
> But it is then better to install the npm: equivalent anyway.

But then what happens when something requires a cross repository dependency?

lookfirst:
In the case of require('foo.json') I still don't get why you can't rewrite that to require('foo.json!systemjs-json')

guybedford:
Generally something written for npm will expect its dependencies on npm, and it's usually best to leave it that way. Consider what the person writing the package would consider the main place they are publishing the package, then you probably want to be using that.

guybedford:
Because it's not just one rule. The rules add up, the cases add up. We need to have crystal clear conventions to avoid even more confusion.

guybedford:
It's a great point though that none of this is clear to users. I've created https://github.com/jspm/github/issues/62 which should hopefully catch them in the act and provide the right guidance here.

lookfirst:
This is all awkward and yucky... now there is a dependency in config.js for moment on `process` and I have to go update my already existing github dependency on moment.

lookfirst:
I give up... npm installed it is...

PR #494, opened 2015-07-23
Create angularjs-slider@0.1.22.json

lookfirst:


lookfirst:
sorry about that... the same exact file worked fine locally with the extra comma in it. @guybedford maybe improve the `-o` system to error in the same way?

PR #507, opened 2015-07-30
Fix jquery sizzle path

nicklasb:
The current path seems to point towards a non-existing folder?
Seems like a to big an error to go unnoticed, or am I misconfigured back here?


guybedford:
Nice, thanks.

PR #510, opened 2015-08-01
Add bootstrap3-dialog, Angular Schema Form ecosystem

nicklasb:
This adds nakupanda/bootstrap3-dialog to the registry.


nicklasb:
Adding ASF ecosystem as well.

PR #512, opened 2015-08-02
socket.io-client is now in socketio, not Automattic.

simontaylor81:
Socket.io seem to have changed their github org name to socketio. Github redirects requests so install still works with the old name, but the override is not applied, so the library won't actually load.

This change adds the additional override for the new org name, and updates the registry to point to it.


guybedford:
Great, thanks.

PR #521, opened 2015-08-12
Add angular-translate-loader-partial registry and override

smajl:


package-overrides/github/angular-translate/bower-angular-translate-loader-partial@2.7.2.json
```diff
@@ -0,0 +1,16 @@
+{
+  "main": "angular-translate-loader-partial",
+  "registry": "jspm",
+  "format": "global",
+  "dependencies": {
+    "angular-translate": "^2.5.2"
```
guybedford: Should probably add `angular` as a dependency as well.

smajl: Yeah, true. To be honest, in this case, I just copied the override from existing `angular-translate-loader-static-files` - changing just few names. :) I added the `angular` dependency.

PR #522, opened 2015-08-14
Add `format` to angular-ui-router

whitneyit:
As you can see in [angular-ui-router.js](https://github.com/angular-ui/ui-router/blob/805e69bae319e922e4d3265b7ef565058aaff850/release/angular-ui-router.js#L10), `module.exports` is being used to define the module name for angular.

Setting the format here allows for the following.

```js
import angular from 'angular';
import uiRouter from 'angular-ui-router'

angular.module('app', [uiRouter]);
```

### DIFF
--- a/package-overrides/github/angular-ui/ui-router@0.2.12.json
+++ b/package-overrides/github/angular-ui/ui-router@0.2.12.json
@@ -2,6 +2,7 @@
   "directories": {
     "lib": "release"
   },
+  "format": "cjs",
   "main": "angular-ui-router",
   "shim": {
     "angular-ui-router": {


whitneyit:
After removing my `jspm_packages` folder and reinstalling I see this

```sh
$ cat ~/.jspm/jspm-cache/package-overrides/github/angular-ui/ui-router\@0.2.12.json
```
```json
{
  "directories": {
    "lib": "release"
  },
  "format": "cjs",
  "main": "angular-ui-router",
  "shim": {
    "angular-ui-router": {
      "deps": ["angular"]
    }
  },
  "registry": "jspm",
  "dependencies": {
    "angular": "^1.3.2"
  }
}
```

However the file shows the following.

```sh
$ head jspm_packages/github/angular-ui/ui-router\@0.2.15/angular-ui-router.js
```
```js
/* */
"format global";
"deps angular";
/**
 * State-based routing for AngularJS
 * @version v0.2.15
 * @link http://angular-ui.github.com/
 * @license MIT License, http://www.opensource.org/licenses/MIT
 */
```

Is this correct? The `"format global"` is casuing problems still.

whitneyit:
To clarify this only happens when using `System.import('src/app')`.

It works fine with `build` and `buildSFX`

guybedford:
How is the `format global` causing problems? Shims always only work on globals with the current config.

whitneyit:
When I change it to `format cjs` it works. I have no idea why. I made a [pen](http://codepen.io/anon/pen/JdqvLV?editors=100) to demonstrate the problem yet it worked perfectly :confused:

guybedford:
Yes, if you can isolate this further that would really help to consider.

PR #530, opened 2015-08-18
Scope materialize to the dist directory

adamnfish:
The entire materialize repo is nearly 30MB, this change means we only put the relevant parts into `jspm_packages`.

### DIFF
--- a/package-overrides/github/Dogfalo/materialize@0.97.0.json
+++ b/package-overrides/github/Dogfalo/materialize@0.97.0.json
@@ -1,4 +1,7 @@
 {
+  "files": [
+    "dist"
+  ],
   "main": "dist/js/materialize",
   "shim": {
     "dist/js/materialize": {


guybedford:
Thanks, it's best not to alter an override `directories.lib` as that is a breaking change, so it would be best to just use the `files` and `ignore` properties here if possible.

adamnfish:
Whatever you suggest, but having the default entry in the jspm registry set up so you end up hosting 30MB of dev cruft seems to be to be a mistake that needs fixing.

This override seems to be working well for me, so I thought it was safe to propose it. A combination of files and ignore wouldn't be very practical because there are a large number of files that need to be included from various directories (which happens to be precisely the contents of the `dist` directory) or an even more massive amount of cruft that would need to excluded.

[Check it out](https://github.com/Dogfalo/materialize), you'll see what I mean.

adamnfish:
I've had another crack using `files` and cleaned the history on this branch. Let me know if this looks better?

(is it a bug that jspm still includes all the empty directories when using `files`?)

guybedford:
Thanks, that looks great. Yes that is a bug with files, see https://github.com/jspm/jspm-cli/issues/983.

PR #533, opened 2015-08-20
howler.js@2.0

mmoss:
* Only includes the core package since it's stable in the 2.0 branch.
* **plugins** package to be added once stable.

Tested with:
```
npm run jspm -- i howler=github:goldfire/howler.js@2.0 -o "{ 'main': 'src/howler.core.js', 'files': [ 'src/howler.core.js' ] }"
```


guybedford:
Thanks for posting. It's usually best to use the unminified version as it provides a better debugging experience. Let me know if you're able to update that here?

mmoss:
@guybedford Agreed. It was late when I did this... updated to point to the proper source file now. Thanks!

PR #534, opened 2015-08-20
bootstrap@4.0.0-alpha

jdanyow:


package-overrides/github/twbs/bootstrap@4.0.0-alpha.json
```diff
@@ -0,0 +1,25 @@
+{
+  "main": "js/bootstrap",
+  "files": [
+    "dist/js",
+    "dist/css"
+  ],
+  "ignore": [
+    "dist/js/umd",
```
guybedford: Perhaps we should keep these available for modular use?

package-overrides/github/twbs/bootstrap@4.0.0-alpha.json
```diff
@@ -0,0 +1,25 @@
+{
+  "main": "js/bootstrap",
+  "files": [
+    "dist/js",
+    "dist/css"
+  ],
```
guybedford: I'm wondering if we don't just remove this entirely and set `files: null`.

guybedford:
Thanks so much for putting this together! Note that this override is having to "cancel out" the override already present in https://github.com/twbs/bootstrap/blob/master/package.js. If you're able to provide a corresponding PR to bootstrap that would be a great help, otherwise I'll go ahead and do this when we merge.

package-overrides/github/twbs/bootstrap@4.0.0-alpha.json
jdanyow: sounds good

package-overrides/github/twbs/bootstrap@4.0.0-alpha.json
jdanyow: ok will give that a try

jdanyow:
I can create a PR for the bootstrap repo- I should be modifying this file right:
https://github.com/twbs/bootstrap/blob/v4-dev/package.json#L83

(you linked to their `package.js` in master above- I think that was a typo but just wanted to make sure).

guybedford:
Thanks, yes exactly, it should be on the 4 dev branch. Just let me know if you have any further questions at all here.

jdanyow:
I've made the changes you mentioned and created a PR in the bootstrap v4 alpha branch.  I removed the `"files": null` in the bootstrap repo PR.

Let me know if any other tweaks are required.

Thanks!

guybedford:
Great thanks for the update.

PR #536, opened 2015-08-21
Added jquery.carouFredSel and overrides

gig177:


package-overrides/github/gilbitron/carouFredSel@6.2.1.json
```diff
@@ -0,0 +1,8 @@
+{
+    "main": "jquery.carouFredSel*",
```
guybedford: Wildcards aren't supported in mains - it would just be using the bower main in this case.

PR #538, opened 2015-08-23
Add Redux and React Redux to registry

jackfranklin:
[Redux](https://github.com/rackt/redux)
[React-Redux](https://github.com/rackt/react-redux)

@guybedford what do you think?


guybedford:
Looks great to me. Are these working ok in jspm for you?

jackfranklin:
@guybedford so far so good, although still very early days.

Issue #543, opened 2015-08-25
moment locale loading fails


DominicBoettger:
Uncaught TypeError: Cannot read property 'defineLocale' of undefined

import 'moment';
import 'moment/locale/de';

Thought the overrides fixes this. A jspm install moment installed moment 2.10.6.

DominicBoettger:
I tried this, but it did not work.

```
"github:moment/moment@2.10.6": {
        "shim": {
          "moment/locale/*.js": {
            "deps": [
              "moment"
            ]
          }
        }
      }

cAhmad:
I got the same issue using the `shim` as well but with the `npm` version

guybedford:
Which version of jspm are you running? In 0.17 beta `jspm install moment && jspm run moment/locale/de -v` seems to work fine to me.

cAhmad:
Sorry for the disturb, it was my mistake: `jspm install moment` just work

PR #544, opened 2015-08-26
fixed moment locales

DominicBoettger:
Fixed the issue #543


guybedford:
Thanks!

guybedford:
Actually, sorry one thing, we should change the `deps` value to `../moment` in case users install moment under a name that is not `moment` (portability and all).

PR #548, opened 2015-08-27
Add aurelia-i18n

zewa666:
Adds registry info for aurelia's internationalization plugin


zewa666:
you are freaking fast @guybedford :+1:

PR #550, opened 2015-08-27
Add mbe-nodes and angular-ui-layout

nicklasb:
This adds the frontend for MBE and the angular-ui-layout package that it depends upon.

### DIFF
--- a/package-overrides/github/angular-ui/ui-layout@1.1.0.json
+++ b/package-overrides/github/angular-ui/ui-layout@1.1.0.json
@@ -1,7 +1,7 @@
 {
   "format": "global",
   "dependencies": {
-    "angular": "^1.0.2"
+    "angular": "^1"
   },
   "registry": "jspm",
   "main":
--- a/package-overrides/github/angular-ui/ui-layout@1.1.0.json
+++ b/package-overrides/github/angular-ui/ui-layout@1.1.0.json
@@ -1,15 +1,13 @@
 {
-  "format": "global",
   "dependencies": {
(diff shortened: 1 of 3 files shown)


package-overrides/github/angular-ui/ui-layout@1.1.0.json
```diff
@@ -0,0 +1,15 @@
+{
+  "format": "global",
+  "dependencies": {
+    "angular": "^1"
+  },
+  "registry": "jspm",
+  "main":
+    ["./src/ui-layout.js"],
```
guybedford: jspm doesn't stricly support a main array, may be best to make this a plain string.

package-overrides/github/angular-ui/ui-layout@1.1.0.json
```diff
@@ -0,0 +1,15 @@
...
+  "shim": {
+    "ui-utils": {
+      "deps": "angular"
+    }
+  },
+  "format":"amd"
```
guybedford: Is it global or AMD here?

guybedford:
Looks great, thanks... just let me know about the minor points above.

nicklasb:
1. I'll change that into a string.
2. Oh. That was certainly strange of me to miss, I'll have to blame that on the sun spots. I am even not sure either is needed anymore, I'll get back to you on that.

nicklasb:
The format doesn't seem to be needed with the new version.

guybedford:
:+1:

PR #557, opened 2015-09-03
Added jquery.dotdotdot

gig177:


package-overrides/github/BeSite/jQuery.dotdotdot@1.7.4.json
```diff
@@ -0,0 +1,16 @@
+{
+    "main": "jquery.dotdotdot",
+    "deps": [
```
guybedford: This should be a `dependenices` object. See other overrides for examples.

gig177:
How should i specify install jquery too?

guybedford:
See for example https://github.com/jspm/registry/commit/a3efb64471c6a1a1c3a01ec1a1f1003b5b86b60c#diff-50bfc48565a6e99b030a3cafe14adecaR18.

gig177:
Thanks.

package-overrides/github/BeSite/jQuery.dotdotdot@1.7.4.json
```diff
@@ -0,0 +1,16 @@
...
+        "jquery": "*"
+    },
+    "shim": {
+        "jquery.dotdotdot": {
+            "deps": [
+                "jquery"
+            ]
+        }
+    },
+    "directories": {
```
guybedford: We don't encourage `directories.lib` as there is a chance a user may have already installed this and be relying on the existing paths. Best to include `src/js` in all paths for compatibility. The only time we can safely set `directories.lib` is if this is checked into the original package.json file as a PR and bumped into a major release.

PR #566, opened 2015-09-10
typescript: add overrides for npm modules not used when 'browser' is …

frankwallis:
…defined

when 'browser' is defined typescript skips requiring any of these modules. See https://github.com/jspm/jspm-cli/issues/1092

also added override file for current overnight build.


guybedford:
:+1:

PR #570, opened 2015-09-17
Add big.js override

JeroenVinke:
it says [here](https://github.com/jspm/registry/wiki/Configuring-Packages-for-jspm#shim) that exports are detected automatically, but this didn't work for big.js. This fixed it: `jspm install big.js -o "{ shim: { big: { exports: 'Big' } } }"`. please verify if I've created the override correctly

### DIFF
--- a/package-overrides/npm/big.js@3.1.3.json
+++ b/package-overrides/npm/big.js@3.1.3.json
@@ -1,0 +1,7 @@
+{
+  "shim": {
+    "big": {
+      "exports": "Big"
+    }
+  }
+}


guybedford:
@JeroenVinke can you clarify what you mean that the exports aren't detected automatically? This library seems to install and load fine for me already?

JeroenVinke:
When using big.js without the override, I get the error `Uncaught ReferenceError: Big is not defined`. I also get this error when using it from the chrome console. I have verified that big.js is indeed loaded. No issues when I add the override though

guybedford:
@JeroenVinke is this specifically when running in Node or the browser?

If I do -

```javascript
System.import('big.js').then(function(Big) {
  console.log(new Big(5));
});
```

it doesn't throw any errors for me.

JeroenVinke:
You're right, nothing wrong here. Adding `import 'big.js';` fixed it. Pretty important detail.

guybedford:
:+1:

PR #572, opened 2015-09-17
feat(registry): add cloudinary to registry

aaronroberson:
Add cloudinary_js to registry and include overrides for dependent files

### DIFF
--- a/package-overrides/github/cloudinary/cloudinary_js@1.0.25.json
+++ b/package-overrides/github/cloudinary/cloudinary_js@1.0.25.json
@@ -1,4 +1,4 @@
-"github:cloudinary/cloudinary_js@1.0.25": {
+{
   "main": "js/jquery.cloudinary",
   "map": {
     "jquery.ui.widget": "./js/jquery.ui.widget"
--- a/package-overrides/github/cloudinary/cloudinary_js@1.0.25.json
+++ b/package-overrides/github/cloudinary/cloudinary_js@1.0.25.json
@@ -1,0 +1,16 @@
+"github:cloudinary/cloudinary_js@1.0.25": {
+  "main": "js/jquery.cloudinary",
+  "map": {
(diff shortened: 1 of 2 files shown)


guybedford:
A discussion came up recently whether we should be using Angular from npm or the GitHub ecosystem. I'm wondering if it's worth considering again if we should use Angular from npm in https://github.com/jspm/jspm-cli/issues/1132. It would likely involve some breaking registry changes at this point unfortunately. Any feedback you have on that is welcome.

aaronroberson:
I've used the npm versions of Angular with Webpack and have had great success. My biggest concern with updating the jspm registry with the npm version of Angular is that many of the github packages that depend on Angular will end up sticking angular in the github packages in addition to the npm package. This has somehow caused Angular to load twice and providing an appropriate warning. I imagine a map override would have to be created for every Angular module installed from github to point Angular at the npm version which could be cumbersome.

Also, I'm wondering why I can only import packages, as opposed to being able to import and require them? ES6 allows for both, and even more goodness (please see http://www.2ality.com/2014/09/es6-modules-final.html) but my jspm app only supports import.

package-overrides/github/cloudinary/cloudinary_js@1.0.25.json
```diff
@@ -0,0 +1,16 @@
+{
+  "main": "js/jquery.cloudinary",
+  "map": {
+    "jquery.ui.widget": "./js/jquery.ui.widget"
+  },
```
guybedford: I don't think this map is necessary? Or is there an AMD file in this repo that can be loaded?

guybedford:
See the one comment above. I'm not sure I understand what you mean about not being able to require packages, can you clarify?

aaronroberson:
The jquery.fileupload.js does an AMD require of jquery.ui.widget.js so I had add the map so that it does the lookup from the same folder and not from the app root.

guybedford:
:+1:

PR #573, opened 2015-09-18
Add angular-translate-once and angular-translate-handler-log.

vigie:
Add package override for angular-ui-extras to configure the angular dependency
and restrict the downloaded files.


guybedford:
A discussion came up recently whether we should be using Angular from npm or the GitHub ecosystem. I'm wondering if it's worth considering again if we should use Angular from npm in https://github.com/jspm/jspm-cli/issues/1132. It would likely involve some breaking registry changes at this point unfortunately. Any feedback you have on that is welcome.

package-overrides/github/christopherthielen/ui-router-extras@0.0.14.json
```diff
@@ -0,0 +1,15 @@
+{
+  "files": [
```
guybedford: Let's leave this out, in case usage of the modular versions is useful for others in future.

guybedford:
This looks great, see the one comment above. Apologies for the delay on this, just wanted to ensure we're on the right track with using GitHub for these.

package-overrides/github/christopherthielen/ui-router-extras@0.0.14.json
vigie: Done.

PR #574, opened 2015-09-19
add overrides for typescript 1.6.2

frankwallis:
identical to 1.6.0-beta...


guybedford:
Thanks, this shouldn't need to be added though as its in the semver range of the 1.6.0-beta override? Is it definitely not applying?

frankwallis:
no it's definitely not applying

guybedford:
Thanks, have noted at https://github.com/jspm/jspm-cli/issues/1101.

PR #576, opened 2015-09-21
Add toastr

32graham:


package-overrides/github/CodeSeven/toastr@2.1.1.json
```diff
@@ -0,0 +1,22 @@
...
+  "shim": {
+    "toastr": {
+      "deps": [
+        "jquery",
+        "./toastr.css!"
+      ],
+      "exports": "toastr"
+    }
+  }
+},
```
guybedford: Trailing comma.

package-overrides/github/CodeSeven/toastr@2.1.1.json
```diff
@@ -0,0 +1,22 @@
+{
+  "main": "toastr",
+  "registry": "github",
+  "format": "global",
+  "dependencies": {
+    "jquery": "github:components/jquery@2.1.3",
+    "css": "github:systemjs/plugin-css@^0.1.13"
```
guybedford: Best to leave these as the widest ranges possible - `jquery: 'github:components/jquery'` (meaning `*`), and `*` for CSS as well.

guybedford:
Perhaps the version on `npm` is a smaller download?

32graham:
I just noticed there is another PR #480 that is doing the same thing (whoops my bad). I will continue cleaning up this PR until told otherwise.

32graham:
@guybedford the npm download is 3.49 MB and the GitHub download is 572 KB for the latest version (v2.1.2). So I guess we should stick with GitHub in this case.

I could not get it working with just `*` but I was able to get it working without the version numbers for jquery and plugin-css.

I also had to change where I grab the css from so that the same override would work for v2.1.1 and v2.1.2.

Anything else for me to fix here?

package-overrides/github/CodeSeven/toastr@2.1.1.json
```diff
@@ -0,0 +1,22 @@
...
+  },
+  "files": [
+    "build/toastr.min.css",
+    "toastr"
+  ],
+  "shim": {
+    "toastr": {
+      "deps": [
+        "jquery",
+        "./build/toastr.min.css!"
```
guybedford: Can we use the unminified version here?

package-overrides/github/CodeSeven/toastr@2.1.1.json
```diff
@@ -0,0 +1,22 @@
...
+  "registry": "github",
+  "format": "global",
+  "dependencies": {
+    "jquery": "github:components/jquery",
+    "css": "github:systemjs/plugin-css"
+  },
+  "files": [
+    "build/toastr.min.css",
+    "toastr"
+  ],
```
guybedford: Perhaps make this `["toastr.js", "build"]`? Also `toastr` without an extension in theory shouldn't work here.

package-overrides/github/CodeSeven/toastr@2.1.1.json
```diff
@@ -0,0 +1,22 @@
+{
+  "main": "toastr",
+  "registry": "github",
```
guybedford: This is unnecessary and implied by default for anything on GitHub.

guybedford:
Thanks, see comments above. This is looking great, and agreed the GitHub version is best. We should also send a PR to the original project to get this config in there too.

32graham:
@guybedford thanks for your patience as I figure out this process. After I fully understand it, I should be able to help out with getting these config's into popular libraries.

I have split v2.1.1 and v2.1.2 into two separate configs because the non-minified css is not in the same place for both releases.

Should I include the entire "build" directory if I am not using all those files? Should .min and .map files be included in addition to the non-minified versions?

Suppose I get this config right, then get a pull request accepted into the original repo... should I come back here and remove the config from the jspm/registry repo?

guybedford:
It can be useful to include all the build files in case other users find that beneficial. There is no harm leaving extra files in as it does not affect the download time, only the build time.

This looks great to merge though, thanks for your patience on the responses.

guybedford:
(Just let me know if you're planning to update the `files` array or not, so I can merge when its ready)

32graham:
I will update it to include the build folder. It will be next week until I get to it though.

32graham:
@guybedford I now consider this PR ready if you do.

32graham:
Also, @guybedford if could take a peek at CodeSeven/toastr#352 to make sure it looks correct that would be great :)

guybedford:
The PR there looks great. As soon as that is in, we can include an _empty_ override here in the registry for that published version to indicate the shift. Thanks so much for your work on this, really appreciated.

PR #581, opened 2015-09-22
xlsx@0.8.0

teohhanhui:


guybedford:
We actually have an override at https://github.com/jspm/registry/blob/master/package-overrides/github/SheetJS/js-xlsx%400.8.0.json already. Does that version not work for you here?

teohhanhui:
Sorry, I did not notice that as I was using `npm` registry (why not?).

That override seems incomplete. It's not loading `jszip`.

guybedford:
The version of xlsx from `npm` will install all of its dependencies. Are they all necessary for execution of the project? I'll gladly make the npm version the default if it is better configured? Then we should also alter the entries at https://github.com/jspm/registry/blob/master/registry.json#L431 as well.

//cc @tauren

teohhanhui:
It doesn't work if `jszip`\* is not loaded (and exposed as a global). I wouldn't have bothered looking into creating this override otherwise.

\* But it's using bundled `jszip.js` instead of `jszip` package stated in package.json dependencies. Is there a way to skip the installation? (Not really a problem though...)

guybedford:
You can skip dependency installs by mapping them to `@empty` - `jspm install x -o "{map: {somedep: '@empty'}}"`. 

The dependency overhead is certainly a concern when configuring the override. And the npm version has a lot of stuff going on.

I can only merge this if you're able to put in the leg work to justify the override.

teohhanhui:
Updated. None of the dependencies specified in `package.json` are needed/used in a browser environment (`codepage` is optional).

guybedford:
Excellent thanks! Shall we make the `npm` the main one then and update https://github.com/jspm/registry/blob/master/registry.json#L431 to reflect this?

teohhanhui:
I don't see the benefit of using the `npm` version. There doesn't seem to be any external dependency that needs to be loaded.

(I'm not sure about `ods.js` because I don't use that... Does `"format": "global"` affect all files in the package? If that's the case then it should work fine with `ODS` exposed as a global.)

guybedford:
Sure, then let's leave out the `npm` override altogether and just update the github one here?

teohhanhui:
Actually, what's the recommended registry to use? Are users expected to always try `jspm` before anything else? I think it'd be good if we can have the same override for `npm` in this case, since that's the de facto go-to registry for most people. Seems harmless to me...

guybedford:
@teohhanhui it's important to choose a registry as the main registry for each project. Overrides should then only be present for that version. This is in order to ensure we don't have conflicts where one dependency uses one registry and another dependency uses a different registry of the same project, which can't be resolved.

teohhanhui:
I've updated accordingly. But it seems apparent that we need to be able to redirect a registry lookup for a certain package to their counterpart in a different registry (e.g. `npm:xlsx` -> `github:SheetJS/js-xlsx`). Otherwise there will be more broken installations like what I've experienced...

guybedford:
Great, thanks. Redirects are supported, but it can be desireable to install from a specific registry when testing out alternatives. We could warn when installing a project if it is in the registry or npm as preferable though and develop workflows around that, but it's a little work and difficult to tell how much users would respect that. Encouraging `jspm install x` before `jspm install npm:x` is the current focus.

PR #591, opened 2015-09-30
react-transform-catch-errors stuff

timbur:
It took entirely too long to figure out the right combination of overrides to make these 3 packages work together via `jspm`, so I'd like to have this in the registry for future use.

If/when @gaearon accepts https://github.com/gaearon/babel-plugin-react-transform/pull/28 you can then point `babel-plugin-react-transform` to his repo instead of mine.


gaearon:
Anything I can make easier for you on my end?

timbur:
@gaearon I think the root of the problem is with the double interop.  The only thing I can think of that you could do to make things a bit easier in this regard (until https://github.com/babel/babel/issues/2148 is sorted out, at least) is to no longer `rimraf` the `src` directory so that I can `jspm install npm:babel-plugin-react-transform@^2.0.0 -o '{registry: "npm", main: "src/index"}'` (and similarly for others like `react-transform-catch-errors`) instead of the currently super long `jspm install babel-plugin-react-transform=github:gaearon/babel-plugin-react-transform@^2.0.0 -o '{registry: "npm", main: "src/index"}`.

package-overrides/npm/redbox-react@1.1.1.json
```diff
@@ -0,0 +1,9 @@
+{
+  "registry": "npm",
```
guybedford: This is assumed by default for all packages on npm so is unnecessary.

package-overrides/npm/redbox-react@1.1.1.json
```diff
@@ -0,0 +1,9 @@
...
+  "main": "dist/redbox",
```
guybedford: What was the problem with using the `lib/index` here? It seems like dist builds in `object-assign` and `error-stack-parser` right? Do these not work well as separate dependencies?

package-overrides/npm/redbox-react@1.1.1.json
```diff
@@ -0,0 +1,9 @@
...
+  "format": "cjs",
```
guybedford: This is assumed by default for packages on npm.

package-overrides/github/gaearon/react-transform-catch-errors@1.0.0.json
```diff
@@ -0,0 +1,5 @@
+{
+  "registry": "npm",
```
guybedford: What is the reason for not using `npm:react-transform-catch-errors` here?

package-overrides/github/gaearon/react-transform-catch-errors@1.0.0.json
```diff
@@ -0,0 +1,5 @@
...
+  "main": "src/index",
+  "format": "es6"
```
guybedford: Was there a reason you want to use the ES6 sources over the CommonJS transpiled sources?

package-overrides/github/timbur/babel-plugin-react-transform@2.0.0.json
```diff
@@ -0,0 +1,5 @@
+{
+  "registry": "npm",
+  "main": "src/index",
+  "format": "es6"
```
guybedford: Same questions as above?

guybedford:
Thanks @timbur for your work on this, I know overrides can be some effort! Just let me know if I can help further.

package-overrides/npm/redbox-react@1.1.1.json
gaearon: I think `dist/redbox` might have been from some earlier configs circling around where Browserify/Webpack build had issues. I believe those were fixed since then but I might be wrong.

package-overrides/github/gaearon/react-transform-catch-errors@1.0.0.json
timbur: The only way I could prevent the double interop from breaking everything was to use the original ES6 sources.  As for not using `npm:react-transform-catch-errors`, that was because @gaearon removes the `src` directory when publishing to npm, which is of course needed for the `ES6` code.

guybedford: This double interop is designed to work - what exactly breaks?

package-overrides/npm/redbox-react@1.1.1.json
timbur: That's what I thought, but it wasn't installing the dependencies via npm without setting the registry to npm.

package-overrides/github/gaearon/react-transform-catch-errors@1.0.0.json
gaearon: I have no qualms about leaving `src` in the NPM module if that helps you folks.

package-overrides/npm/redbox-react@1.1.1.json
timbur: Specifying `dist/redbox` and the format as `cjs` was the only way I could make this work.  I don't remember the error messages.  I can try to reproduce, step-by-step, all the errors I ran into when trying to install and use these 3 packages.

Actually, I went ahead and put together a repro repo: https://github.com/timbur/another-systemjs-test

How I reached that point:

1.  `npm install express --save`
2.  `npm install jspm --save-dev`
3.  `jspm init` -> Babel -> etc.
4.  `jspm install npm:babel-plugin-react-transform npm:react@^0.14.0-rc1 npm:react-dom npm:react-transform-catch-errors npm:redbox-react`
5.  Create `config.dev.js` and `reporterOptions.js` and add `transform` to `babelOptions`.
6.  Create `index.html` and `server.js` using `express` static files.
7.  Create `mount.js` and `src/components/App.js`.
8.  Open http://localhost:8080 and get an error that eventually boils down to the double interop, or at least that's what I think it is.  Maybe I'm just being stupid.  But ideally, I think everything I've done here should have worked without any extra effort or debugging.

timbur: I created a branch with the overrides so you can see that they allow the app to correctly load: https://github.com/timbur/another-systemjs-test/tree/using-overrides

guybedford: @timbur thanks for sharing. I can confirm that the example works fine if I:

* Remove the `npm:redbox-react` override entirely and install it fresh
* Install `jspm install npm:react-transform-catch-errors` instead of `github:gaearon/react-transform-catch-errors`
* Simplifying the override for `github:timbur/babel-plugin-react-transform` to just `{ main: 'src/index' }` was all that was necessary to make everything work.

In the above I didn't hit on any double interop issue case - if you can share what that is I'd be grateful.

timbur: @guybedford Did you try both branches?  From your description, it looks like you didn't try the master branch (which is where the double interop issue is) since you mentioned an override.  The master branch has no overrides.  You might also want to try the 8 steps I listed for creating the master branch.

guybedford: @timbur the error I get on master is `Uncaught (in promise) Error: imports[1] for react-transform-catch-errors does not look like a React component.` is that the one you mean?

timbur: @guybedford Yes.  IIRC, I tried logging `imports[1]` to see what it was, and it was ultimately `{ default: etc }`.

guybedford: Ok right I see. So it's loading the module and expecting it to be the full module. This is actually not an interop bug, but the way that ES modules work - they are always namespace objects when loaded unlike CommonJS. The fact that the module is loaded at all is actually pretty nice to see despite the code being designed for Node! Perhaps a simple check like `module.default || module` check within `react-transform-catch-errors` is what is needed here?

timbur: What I'm personally trying to avoid is having to fork and/or create PRs for every case like this in the future.  If this is unavoidable due to the way ES modules work, then I guess from now on, for every transpiled ES6 module, I'll just have to add the `{format: "es6", main: "src/index"}' overrides?  Although, I'm curious as to why I've never run into this problem with `webpack`, no matter what combination of `cjs` or `es6` modules, Babel plugins, etc.

guybedford: How exactly is `imports[1]` being loaded above and what is it? I couldn't actually quite work that out.

From what I can tell, this is because `react-transform-catch-errors` expects module objects to be components? While ES modules are always namespaces. 

If you want everything to work without effort, I'd advise not using jspm.

timbur: `imports[1]` in this case is `redbox-react`.  It gets passed to `react-transform-catch-errors` to report errors to the developer.  I'm not so much using `jspm` as I am `systemjs`, and I'm using it as a way to dynamically load arbitrary modules within the browser, without the need for a server + webpack.

guybedford: Ah, ok. Right I see the issue you're having. I've created https://github.com/gaearon/babel-plugin-react-transform/issues/35 as this is a babel-plugin-react-transform issue.

guybedford: @timbur because most projects out there don't support SystemJS and jspm, the project does require more time investment to make things work. We've done a lot to try and support existing things, but there will always be edge cases that make it a bumpy ride. Checking in overrides, posting issues and project maintainers checking their projects against jspm all contribute to improving this, which will take some time to match the development experience in other systems.

PR #595, opened 2015-10-02
Update registry.json

veggiemonk:
add tape (for testing js)
https://github.com/substack/tape


guybedford:
Thanks, is `npm:tape` not preferable here?

veggiemonk:
You are right.

guybedford:
Thanks!

PR #596, opened 2015-10-03
add @reactivex/rxjs

robwormald:
adds @reactivex/rxjs to registry.

### DIFF
--- a/registry.json
+++ b/registry.json
@@ -439,5 +439,6 @@
   "zlib": "github:jspm/nodelibs-zlib",
   "zone.js": "npm:zone.js",
   "zygo": "github:zygo-io/zygo",
-  "zygo-blog": "github:zygo-io/zygo-blog"
+  "zygo-blog": "github:zygo-io/zygo-blog",
+  "@reactivex/rxjs": "npm:@reactivex/rxjs"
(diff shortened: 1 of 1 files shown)


robwormald:
@guybedford not sure how this is going to go with the namespacing - also, can't add an override for it due to the @in the name.

guybedford:
@robwormald `@` in the name is fine, but are you sure you don't want to call it `rxjs`?

robwormald:
`@reactivex/rxjs` is how its published on NPM, and how we refer to it inside NG2, so i think so, yes.

robwormald:
ah, the problem with the override is the slash in the name (silly...). suggestions?

guybedford:
@robwormald unfortunately yes I remember now `/` in the name can't actually be allowed for the jspm registry itself. Is this to make sure that angular2 install finds it? If so one can add `npm:@reactivex/rxjs` as the dependency in the package.json to reference the npm version directly. Also I saw that other override fix you tweeted, should we check that in as an override for `npm:@reactivex/rxjs`?

robwormald:
scratch this, we're bringing rx in core in the next bump. so never mind.

PR #597, opened 2015-10-03
Add qunitjs

rjbultitude:
Add a reference to the Qunit JS library on npm


guybedford:
Thanks, is there a reason you're using this over the `qunit` entry? We can potentially update the existing `qunit` to point to `npm:quint`.

rjbultitude:
Hi @guybedford. You're right, it's not needed. `npm:quit` would be better imho.

guybedford:
Happy to update to npm:qunit, just let me know.

PR #604, opened 2015-10-06
add Microsoft datajs lib

hiddenboox:
Add Microsoft datajs-1.1.3 lib

### DIFF
--- a/registry.json
+++ b/registry.json
@@ -14,7 +14,7 @@
   "less": "github:aaike/jspm-less-plugin",
   "ts": "github:frankwallis/plugin-typescript",
   "sass": "github:screendriver/plugin-sass",
-
+  "datajs": "github:hiddenboox/datajs",
   "ace": "github:ajaxorg/ace-builds",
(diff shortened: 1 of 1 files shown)


guybedford:
Thanks, can you make sure to include this in the correct alphabetical position?

hiddenboox:
@guybedford now should be ok

registry.json
```diff
@@ -177,6 +177,7 @@
   "d3": "github:mbostock/d3",
   "c3": "npm:c3",
   "datatables": "github:DataTables/DataTables",
+  "datajs": "github:hiddenboox/datajs",
```
guybedford: To be pedantic, this should come before datatables.

guybedford:
Not quite yet :)

registry.json
aluanhaddad: @hiddenboox Yeah, just move the entry and this will be good to go

hiddenboox:
Ok, at now i can't, tommorow i will do that

PR #609, opened 2015-10-09
Add Parse JS SDK override to registry

andrewimm:
Ignores the modular pieces of the JS SDK, and only pulls in a pre-compiled browser version.

### DIFF
--- a/package-overrides/npm/parse@1.6.5.json
+++ b/package-overrides/npm/parse@1.6.5.json
@@ -1,0 +1,13 @@
+{
+  "dependencies": {
+    "babel-runtime": "^5.8.20"
+  },
+  "devDependencies": {},
+  "files": [
+    "dist/",
+    "lib/browser",
+    "index.js",
+    "LICENSE",
+    "PATENTS"
+  ]
+}


package-overrides/npm/parse@1.6.5.json
```diff
@@ -0,0 +1,9 @@
+{
+  "dependencies": {},
+  "devDependencies": {},
+  "main": "parse-latest.js",
+  "directories": {
+    "lib": "dist"
+  },
```
guybedford: We try to avoid setting this property as it is a breaking change for existing users.

guybedford:
Thanks, let me know if you can update without the `directories` property.

andrewimm:
For my edification, can you explain why the `directories` entry would be a breaking change?

Seems to me that if `main` is declared to be in a directory with no subfiles, this would be equally as restrictive and breaking.

guybedford:
@andrewimm existing downloads would change their file structure, so if a user had installed and then manually required the right main file, their code would stop working on adding this override.

andrewimm:
Looks like this pulls in the appropriate files, while maintaining the original directory structure.

package-overrides/npm/parse@1.6.5.json
```diff
@@ -0,0 +1,8 @@
+{
+  "dependencies": {
+    "babel-runtime": "^5.8.20"
+  },
+  "devDependencies": {},
+  "main": "dist/parse-latest.js",
+  "files": ["dist/parse-latest.js"]
```
guybedford: It's also advisable to leave this property out in case users want to use other files?

andrewimm:
@guybedford, as a maintainer hopefully you can help us get where we want to be with this change. I was only introduced to JSPM recently, and we want to support JS developers on any build system. Today, it seems like the SDK works fine being pulled in from NPM, but some developers voiced concerns that because we include 3 different builds of our SDK, they were ending up with too much to download.

Assuming JSPM builds are designed for browser environments, developers would either want to build against our compiled version (in `dist`) or our modular version (in `lib/browser`). I have no real context or experience here with JSPM, so which of those two directories should we be focusing on? Or should we include both, default to one with the `main` field, and give developers an opportunity to use the other (which to me seems unlikely).

typhonrt:
Hi Guy / Andrew... 

Thanks for taking a look at this pull request as it's the final step for me to get the Backbone / Parse effort out to the public. 

Andrew, in your latest update is the Babel runtime necessary to include in dependencies since it's already part of the parse-latest.js bundle? IE it seems the runtime is necessary for bundling the ES6 version, but not for consuming the existing ES5 parse-latest.js bundle?

Here is an override that I just tested and seems reasonable:
{ 
   "dependencies": {}, 
   "devDependencies": {}, 
   "files": ["dist/", "LICENSE", "PATENTS"], 
   "main": "dist/parse-latest.js" 
}

andrewimm:
Ah, good point Michael. In the browserified build, there's no need for including dependencies. I guess this all hinges on whether we want the `/lib/browser` version or the `/dist/parse-latest.js` version.

typhonrt:
Hmm... So it is actually useful to access the /lib/browser version when working from ES6 source in some cases. I've just tested a few things against the lib/browser source versus the dist/parse-latest.js bundle. 

@guybedford is there anyway to add multiple entries into registry.json like "parse-browser-es6" or "parse-browser-dist" that both install from npm:parse with different overrides for the "main" field?

guybedford:
@andrewimm thanks for taking the effort to support jspm here.

Just to take a step back a bit, what exactly is the reason for needing an override here at all? Does `jspm install npm:parse` not work correctly?

typhonrt:
@guybedford
No it doesn't as I mentioned in a previous issue (https://github.com/systemjs/systemjs/issues/811) as the actual Parse source distribution has mixed CJS / ES6 imports and is only meant to be consumed by Browserify. IE a two step process which as you mentioned is not supported by SystemJS. It also pulls in a metric ton of unnecessary dependencies and really muddies up the SystemJS config.js file with many Node specific dependencies.

guybedford:
@typhonrt in a simple test here, it seems to load the module ok though? If the problem is downloading unnecessary dependencies, that can be resolved by using overrides to filter out the unnecessary dependencies.

typhonrt:
If one does "jspm install npm:parse" it will indeed seemingly work, but none of the source code is in a usable state for SystemJS. 

Yes, the override gets rid of the dependencies / devDependencies from the npm package. 

Perhaps you can explain the "main" field in the override as there are two consumable points of entry as mentioned. An ES5 bundle with dist/parse-latest.js or unbundled ES6 source code in lib/browser. 

By default the ES5 bundle is perfectly fine to consume with SystemJS... In some cases it may be nice to have the main entry point be lib/browser/Parse.js

typhonrt:
OK, hmm... Things seem to work with the "jspm install npm:parse". This wasn't the case necessarily for the 1.6.3 version when I first tested installing the raw npm:parse package. Granted this pulls in a bunch of dependencies for server / react-native runtimes which aren't needed by the browser version. 

I was on a slightly old (~1 month old SystemJS 18.x / JSPM 0.16.2) build when I previously tested this. 

Other than that it's not clear why things didn't work previously. This override even works fine:

"{ 'dependencies': {}, 'devDependencies': {}, 'files': ['dist/', 'lib/', 'index.js', 'LICENSE', 'PATENTS'] }"

The index.js will load "/lib/browser/Parse.js".

andrewimm:
The ideal behavior in a module-based environment (aka, not bower) is to use the `lib/` versions. That way, modules can be packaged according to whatever manager is doing the bundling. This will probably also make the best debugging experience. If that seems to work properly, it's a superior solution to using the `dist` dir.

We should probably still leverage the override to avoid downloading `lib/node` and `lib/react-native`

typhonrt:
@andrewimm I definitely agree with using the lib/ version as being superior as long as the developer consuming it is ES6 based via SystemJS.  Not everyone will be hence the benefit of somehow also enabling the ES5 bundle as the default. As far as not downloading the lib/node and lib/react-native source this override handles that:
```
{ 
   "dependencies": {}, 
   "devDependencies": {}, 
   "files": ["dist/", 'lib/browser/', 'index.js', "LICENSE", "PATENTS"], 
} 
```

andrewimm:
@typhonrt `lib` does not contain ES6 code. It's all ES5 code using commonjs require statements

typhonrt:
@andrewimm  Gotcha... Yeah. I did a quick "cat" of Parse.js and didn't look at the top of the file. I think we're all good then with the above latest override... :D

typhonrt:
@andrewimm @guybedford Just bumping this issue to see if Andrew can get the final pull request in before EOD as described above (create a new one?). I'm stalled on this in order to get my Backbone+Parse repos out and I'd like to do that this weekend as I'm pretty sure I'm at 100% with my effort that I'd like to open source. This seems to be the proper override for the new Parse SDK:

```
{ 
   "dependencies": {}, 
   "devDependencies": {}, 
   "files": ["dist/", 'lib/browser/', 'index.js', "LICENSE", "PATENTS"], 
} 
```

Cheers.. Thanks guys for taking a look earlier today and rapid discussion! :D

guybedford:
@typhonrt I just tested this and it seems to work for me. I'd advise leaving in the `babel-runtime` dependency though. Are you sure the `xmlhttprequest` dependency is not needed? Perhaps we just need the right override for this dependency as opposed to removing it entirely.

andrewimm:
The `babel-runtime` dependency is necessary for the `lib/browser` build; the `xmlhttprequest` module is not. It should be
```
  "dependencies": {
    "babel-runtime": "^5.8.20"
  },
  "devDependencies": {},
```

I'll update the PR tomorrow morning.

typhonrt:
@andrewimm Excellent let me know when this gets into the registry. Did you get the PR updated? I also happen to be using the Babel runtime as a dependency in my work, so didn't pick up on it being required for the Parse SDK. It turns out that one can store the override in the package.json in the JSPM section. That's where the JSPM CLI stores it, so I was able to finish all of the work over the weekend as I wasn't blocked per se in testing installing my repos with "jspm install" . I'm just writing documentation and final sanity testing, so should be posting the Backbone / Parse repos by Tuesday morning. I can always remove the override in the package.json once the official one hits the registry.  

@guybedford JSPM / SystemJS is magic now that I am using it for library authoring as well. I was able to get three separate repos working together which reference each other in source. Up until I got it all working I thought I would have to use git submodules, :: shudder ::, so this is great! I have some suggestions for updating documentation library authors and for SystemJS Builder as I had to figure out several tricky things which aren't documented for bundling various versions of the combined library. I'll post an issue in the Builder repo.

typhonrt:
@guybedford  just bumping this politely to see if we can get this into the registry today. I'm hours away from releasing my Backbone-ES6 + Parse library. Apologies for the nuisance..

guybedford:
@typhonrt sure, thanks for your help getting this one through!

abhishiv:
Hey guy,

any idea why this is broken over jspm cdn? It seems to pull in the node version and not just dist/parse-latest.js.

Test here - http://s.codepen.io/abhishiv/debug/RrZyWL

typhonrt:
@abhishiv perhaps open a new issue as this is a closed PR. 

I am using the Parse registry override via CLI / command line / package.json and it's working fine. 

I tried to setup a "JSPM CDN test", but also didn't have success trying to import `npm:parse` / `npm:parse@1.6.14`, etc. I suggest confirming that other npm modules can be loaded via your own "JSPM CDN test" and post that code on Github perhaps as it's easier to work with there.

PR #610, opened 2015-10-11
slick carousel

jdanyow:
PR to actual library here:  https://github.com/kenwheeler/slick/pull/1807

### DIFF
--- a/package-overrides/github/kenwheeler/slick@1.5.8.json
+++ b/package-overrides/github/kenwheeler/slick@1.5.8.json
@@ -1,0 +1,18 @@
+{
+  "main": "slick",
+  "files": null,
+  "directories": {
+    "lib": "slick"
+  },
+  "shim": {
+    "slick": {
+      "deps": [
+        "jquery"
+      ],
+      "exports": "$"
+    }
+  },
+  "dependencies": {
+    "jquery": "github:components/jquery"
+  }
+}
--- a/registry.json
+++ b/registry.json
@@ -385,6 +385,7 @@
   "showdown": "github:showdownjs/showdown",
   "sifter": "github:brianreavis/sifter.js",
   "skrollr": "github:Prinzhorn/skrollr",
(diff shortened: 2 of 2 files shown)


package-overrides/github/kenwheeler/slick@1.5.8.json
```diff
@@ -0,0 +1,18 @@
+{
+  "main": "slick",
+  "files": null,
+  "directories": {
+    "lib": "slick"
```
guybedford: It's always best to avoid `directories.lib`.

package-overrides/github/kenwheeler/slick@1.5.8.json
```diff
@@ -0,0 +1,18 @@
...
+  "shim": {
+    "slick": {
+      "deps": [
+        "jquery"
+      ],
+      "exports": "$"
+    }
+  },
+  "dependencies": {
+    "jquery": "github:components/jquery"
```
guybedford: Perhaps add a `registry: 'jspm'` and reference `jquery: '*'`. If we ever move jQuery, it could be useful to be referencing the jspm registry this way.

guybedford:
Thanks! To be even more difficult, I'm wondering if we should use the version on npm rather?

package-overrides/github/kenwheeler/slick@1.5.8.json
guybedford: Or just set `dependencies: { jquery: 'jspm:jquery' }`

jdanyow:
ha no problem.  Initially I was using the npm version of slick but I couldn't figure out how to make the jquery dep work.  Maybe the switch to `dependencies: { jquery: 'jspm:jquery' }` will enable using the npm version of slick

guybedford:
Sure, note that on npm we would need an explicit `registry: 'jspm'` property to make that work I believe. Just let me know if you need further help here at all.

PR #613, opened 2015-10-12
npm:moment-msdate@0.1.0

teohhanhui:
Do we really have to create an override for each and every `moment` plugin?


guybedford:
Let me know if you want to look at setting `npm:moment` as the main entry here? I will wait for your response further here.

teohhanhui:
If we have to provide an override for each `moment` plugin anyway, why not change the main `moment` package to not use `global` format (and just shim where necessary)?

guybedford:
@teohhanhui we shouldn't need to provide an override for each moment plugin as most npm packages are designed to work in jspm. The overrides would only be needed for those that don't work out of the box.

PR #615, opened 2015-10-13
Add override for npm:is-subset@0.1.1

shuaibiyy:


guybedford:
Thanks, this is actually a bug though due to files in:

```javascript
   "files": [
    "/*.js",
    "/License.md",
    "/Readme.md",
    "/module/",
    "/test/"
  ],
```
removing unnecssary files. We're tracking this in https://github.com/jspm/jspm-cli/issues/1031.

Perhaps rather use the override locally in the mean time?

shuaibiyy:
Noted. Thanks.

PR #616, opened 2015-10-13
npm:moment-range@2.0.3

teohhanhui:


guybedford:
@teohhanhui as mentioned, I think it may be beneficial to just use all the moment libraries from npm here. I just tried `jspm install npm:moment-range` and it imported fine for me. If you want to move `jspm install moment` over to `jspm install npm:moment` I do think that is preferable at this point given that we'd need to configure all these plugins otherwise.

PR #617, opened 2015-10-15
npm:js-polyfills@0.1.12

teohhanhui:


guybedford:
@teohhanhui please can you explain why this is necessary? I just tried `jspm install npm:js-polyfills` and that imported fine for me.

teohhanhui:
It imports fine but doesn't work as intended. As you can see these are polyfills which need to be able to modify global.

guybedford:
Thanks for clarifying.

PR #618, opened 2015-10-15
dailymotion/hls.js override

Fr3nzzy:
```v0.1.11``` has wrong ```main``` key in ```package.json```.
```master``` is fine


guybedford:
Thanks, did you try `jspm install npm:hls.js`?

Fr3nzzy:
@guybedford there is no hls.js package in NPM at the moment. 
The only way to install hls.js via NPM is using github:
```npm install --save https://github.com/dailymotion/hls.js```

Should i create an override for NPM too?

guybedford:
Oh right, sorry I missed that! No problem, let's merge this then.

PR #619, opened 2015-10-15
Shim for ismyrnow/Leaflet.groupedlayercontrol

neothemachine:


guybedford:
Nice!

neothemachine:
Hm, maybe I did something wrong, but it doesn't seem to work yet. When I do `System.import("github:ismyrnow/Leaflet.groupedlayercontrol")` it should load the leaflet dependency I put in the shim as well, right? But it doesn't. Just tested in jspm.io console.

guybedford:
The CDN override cache only gets updated periodically - try it again now?

neothemachine:
Works, great! :)

Maybe you should change the README of the registry repo since it says 
"If the configuration is correct, this will be accepted and that will 
instantly enable the functionality on the endpoint."

PR #623, opened 2015-10-16
chore(overrides): angular related overrides for jspm depcache

j-walker23:
This includes all overrides i added for depcache to work. Mostly angular deps, a couple non.

### DIFF
--- a/package-overrides/github/angular-ui/bootstrap-bower@0.14.2.json
+++ b/package-overrides/github/angular-ui/bootstrap-bower@0.14.2.json
@@ -1,0 +1,8 @@
+{
+  "registry": "jspm",
+  "shim": {
+    "ui-bootstrap-tpls": [
+      "angular"
+    ]
+  }
+}
--- a/package-overrides/github/angular-ui/bower-ui-grid@3.0.1.json
+++ b/package-overrides/github/angular-ui/bower-ui-grid@3.0.1.json
@@ -1,3 +1,8 @@
 {
-  "main": "ui-grid"
+  "main": "ui-grid",
+  "shim": {
+    "ui-grid": [
+      "angular"
+    ]
+  }
 }
--- a/package-overrides/npm/angular-dragdrop@1.0.5.json
+++ b/package-overrides/npm/angular-dragdrop@1.0.5.json
@@ -1,0 +1,11 @@
+{
+  "registry": "jspm",
+  "shim": {
+    "src/angular-dragdrop": [
+      "angular"
+    ]
+  },
+  "dependencies": {
+    "angular": "github:angular/bower-angular@*"
+  }
+}
--- a/package-overrides/npm/angular-filters@1.1.2.json
+++ b/package-overrides/npm/angular-filters@1.1.2.json
@@ -1,0 +1,11 @@
+{
+  "registry": "jspm",
+  "shim": {
+    "dist/angular-filters": [
+      "angular"
+    ]
+  },
+  "dependencies": {
+    "angular": "github:angular/bower-angular@*"
+  }
+}
--- a/package-overrides/npm/leaflet.heat@0.1.3.json
+++ b/package-overrides/npm/leaflet.heat@0.1.3.json
@@ -1,0 +1,11 @@
+{
+  "registry": "jspm",
+  "dependencies": {
+    "leaflet": "~0.7.1"
+  },
+  "shim": {
+    "dist/leaflet-heat": [
+      "leaflet"
+    ]
+  }
+}
--- a/package-overrides/npm/leaflet.markercluster@0.4.0.json
+++ b/package-overrides/npm/leaflet.markercluster@0.4.0.json
@@ -1,0 +1,7 @@
+{
+  "shim": {
+    "dist/leaflet.markercluster": [
+      "leaflet"
+    ]
+  }
+}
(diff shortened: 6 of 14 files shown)


package-overrides/github/angular-ui/bootstrap-bower@0.14.2.json
```diff
@@ -0,0 +1,8 @@
+{
+  "registry": "jspm",
```
guybedford: Unfortunately this will break in jspm 0.17 as it will try to read the peerDependencies which are a version range format that is not supported in jspm (currently).

I'd suggest adding a `dependencies: { angular: '*' }` as well here, or leaving this property out entirely to ignore the dependencies completely.

package-overrides/github/angular-ui/bower-ui-grid@3.0.1.json
```diff
@@ -1,3 +1,8 @@
 {
-  "main": "ui-grid"
+  "main": "ui-grid",
+  "shim": {
+    "ui-grid": [
+      "angular"
```
guybedford: May be worth adding a `dependencies: { angular: '*' }` here as well.

guybedford: (at the base-level of course, not in the shim)

package-overrides/npm/angular-dragdrop@1.0.5.json
```diff
@@ -0,0 +1,11 @@
+{
+  "registry": "jspm",
```
guybedford: We have a strange convention with the angular libraries where we typically use github repos over npm installs. The main reason for this is that npm installs do a kind of "conversion" of node packages based on the Node package resolution algorithm and environment needs. This can cause unnecessary deps to be downloaded as well.

Perhaps use the GitHub repo instead here? The override should be much the same.

package-overrides/npm/angular-dragdrop@1.0.5.json
```diff
@@ -0,0 +1,11 @@
...
+  "shim": {
+    "src/angular-dragdrop": [
+      "angular"
+    ]
+  },
+  "dependencies": {
+    "angular": "github:angular/bower-angular@*"
```
guybedford: This should work with `angular: '*'`.

package-overrides/npm/angular-filters@1.1.2.json
```diff
@@ -0,0 +1,11 @@
...
+  "registry": "jspm",
+  "shim": {
+    "dist/angular-filters": [
+      "angular"
+    ]
+  },
+  "dependencies": {
+    "angular": "github:angular/bower-angular@*"
+  }
+}
```
guybedford: Same two points as above.

package-overrides/npm/leaflet.heat@0.1.3.json
```diff
@@ -0,0 +1,11 @@
...
+  "registry": "jspm",
+  "dependencies": {
+    "leaflet": "~0.7.1"
+  },
+  "shim": {
+    "dist/leaflet-heat": [
+      "leaflet"
+    ]
+  }
+}
```
guybedford: All packages up to here have the same points applying.

package-overrides/npm/ng-clip@0.2.6.json
```diff
@@ -0,0 +1,15 @@
+{
+  "registry": "jspm",
+  "directories": {
+    "lib": "src"
+  },
```
guybedford: We discourage the use of `directories.lib` as it can be breaking for existing users of the package.

package-overrides/npm/leaflet.markercluster@0.4.0.json
```diff
@@ -0,0 +1,7 @@
+{
+  "shim": {
+    "dist/leaflet.markercluster": [
+      "leaflet"
+    ]
```
guybedford: Again, worth considering npm v github here. If it's working fine, and there aren't unnecessary installs happening, it's certainly fine to check this in though, but just let me know.

guybedford:
Thanks for posting! It's really great to share these for other users of jspm. Looks like a lot of work there... apologies to add more to it, but we do need to carefully consider these too.

j-walker23:
@guybedford That is great, thanks for taking the time to review it! This was my first real open source PR so its cool to get some feedback.

Spent the last couple weeks 24/7 with jspm, once i realized how freaking cool it seemed compared to my slow boring grunt build system i started a massive refactor. 

Guessing since i have seen your name 1000 times while researching that you are an author of jspm and if so seriously thank you for this. Making frontend fun again. If you don't have one of those gittip things you need one!

I will get all of these fixed up for another review. Quick question, with my app being outdated it had some deps that didn't work in either registry. For those i used [jspm-bower-endpoint](https://www.npmjs.com/package/jspm-bower-endpoint) and it worked well enough. Creates another `bower` reg alongside `npm` and `github`. Would there be any reason for me to get those added as well?

guybedford:
Sure no problem, hope it will be the first of many! If you can share the packages that needed that Bower endpoint that would be useful to hear. I don't generally encourage using the bower endpoint as it ends up doubling up on a lot of things that are already in the github endpoint, without much added value. It would be good though to understand the cases that don't work to consider how to support them.

PR #626, opened 2015-10-20
npm:moment@2.10.6

teohhanhui:
`moment` changed to use `npm:moment` as discussed in #613 and #616

Package override taken from moment/moment#2577

### DIFF
--- a/registry.json
+++ b/registry.json
@@ -304,7 +304,7 @@
   "mocha": "npm:mocha",
   "modernizr": "github:Modernizr/Modernizr",
   "module": "github:jspm/nodelibs-module",
-  "moment": "github:moment/moment",
+  "moment": "npm:moment",
   "moment-json-parser": "npm:moment-json-parser",
   "mongoose": "github:gavinaiken/mongoose-browser",
   "moonridge-client": "github:capaj/Moonridge-client",


package-overrides/npm/moment@2.10.6.json
```diff
@@ -0,0 +1,9 @@
+{
+  "directories": {
+    "lib": "src"
+  },
```
guybedford: We don't encourage the use of `directories.lib`.

package-overrides/npm/moment@2.10.6.json
```diff
@@ -0,0 +1,9 @@
...
+  "main": "moment.js",
+  "ignore": [
+    "./src/test/"
```
guybedford: Moment doesn't actually seem to have a `src` directory?

teohhanhui: https://github.com/moment/moment/tree/2.10.6/src

teohhanhui:
@guybedford Unfortunately I'm having some trouble getting this to work. The `jspm` config in moment's package.json seems to be taking precedence over my local override.

```json
{
  "files": [
    "src"
  ],
  "ignore": [
    "src/test/"
  ],
  "main": "src/moment.js"
}
```
This is my local override but I end up with only `package.json` file in the directory.

teohhanhui:
Uhh I've figured it out. The `src` directory is not part of the `npm` package.

teohhanhui:
As it stands, we cannot use `npm:moment`. Need to find another solution then...

PR #627, opened 2015-10-20
point moment at npm

lookfirst:


guybedford:
This will be handled already fine by the PR at https://github.com/jspm/registry/pull/626 when this lands.

PR #630, opened 2015-10-22
Use npm version of PouchDB

nolanlawson:
See discussion in https://github.com/pouchdb/pouchdb/issues/4460#issuecomment-149931111. It's much less problematic for PouchDB users to use the npm version rather than the Github version.

### DIFF
--- a/registry.json
+++ b/registry.json
@@ -344,7 +344,7 @@
   "pixi.js": "npm:pixi.js",
   "poly": "github:cujojs/poly",
   "postal": "npm:postal",
-  "pouchdb": "github:pouchdb/pouchdb",
+  "pouchdb": "npm:pouchdb",
   "pretender": "npm:pretender",
   "prism": "github:LeaVerou/prism",
   "process": "github:jspm/nodelibs-process",


guybedford:
`jspm install npm:pouchdb` installs a lot of dependencies. Are there features provided by this version that we don't already get in the github version? Ideally overrides can be used to reduce the dependency tree size to just what is necessary here.

nolanlawson:
PouchDB can be installed one of two ways: as a Node lib or as a browser lib. When you `npm install pouchdb`, you get a `lib/index.js` that gives you the Node version, or when browserified gives you the browser version. There's also a prebuilt `dist/pouchdb.js` file that gives you the bundled browser version.

Based on this information, which one is the most appropriate default for `jspm install pouchdb`? I don't know the jspm community very well or what people expect, but based on pouchdb/pouchdb#4460 it seems like folks are getting very confused.

guybedford:
I think the github repo currently may be the best option if it is a browser build and doesn't litter with unnecessary dependencies. What exactly is the issue with using this currently?

nolanlawson:
@guybedford The issue is that it doesn't work out of the box for at least one of our users (see  pouchdb/pouchdb#4460). Looks like the fix for us is to add `{"jspm": {"main": "dist/pouchdb.js"}}`.

Is jspm designed primarily for the browser? If that's the case, then yeah, npm is going to give you a lot of stuff that doesn't help you much (such as a Node version of PouchDB that runs on LevelDB). If that's the case, then I'm fine with us just adding that `package.json` fix and calling it a day.

Will close unless you think there's more that should be done.

guybedford:
@nolanlawson thanks for sorting this out! jspm is designed primarily for the browser today, but we do provide Node workflows as well for universal JS use cases. Happy to stick with what works for now, and then it may be worth revisiting this once jspm 0.17 lands and we can fully support Node execution. As long as you don't mind a bit of back and fourth that is.

Hypercubed:
If `jspm.main` is set to use `dist/pouchdb.js` do you also need to explicitly set the dependencies to `{}`?  Won't all the dependencies also be downloaded if someonbe does `jspm install npm:pouchdb`?

PR #633, opened 2015-10-27
Add ionic2

vadimpopa:
Should that work for ionic2 ?. Also I've tried the github:driftyco/ionic2 and is taking the alpha1 version (github:driftyco/ionic2@2.0.0-alpha.1), but the npm one (npm install ionic@alpha) is taking alpha 30, so preferably to take the npm one

### DIFF
--- a/registry.json
+++ b/registry.json
@@ -234,7 +234,7 @@
   "infamous": "npm:infamous",
   "intro": "github:usablica/intro.js",
   "ionic": "github:driftyco/ionic-bower",
-  "ionic2": "npm:ionic@alpha",
+  "ionic2": "ionic-framework",
   "jcrop": "github:tapmodo/Jcrop",
   "jiff": "github:cujojs/jiff",
   "joi": "github:capaj/joi-browser",
--- a/registry.json
+++ b/registry.json
@@ -234,7 +234,7 @@
   "infamous": "npm:infamous",
   "intro": "github:usablica/intro.js",
   "ionic": "github:driftyco/ionic-bower",
-  "ionic2": "ionic-framework",
+  "ionic2": "npm:ionic-framework",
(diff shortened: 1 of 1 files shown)


registry.json
```diff
@@ -234,6 +234,7 @@
   "infamous": "npm:infamous",
   "intro": "github:usablica/intro.js",
   "ionic": "github:driftyco/ionic-bower",
+  "ionic2": "npm:ionic@alpha",
```
guybedford: Versions aren't supported in the registry.

guybedford:
Sure, does it seem to be working ok for you?

vadimpopa:
"jspm install npm:ionic@alpha" is not working locally. Only "jspm install github:driftyco/ionic2". Then, now I got it right, we should change with the second one and will be fine. So how we proceed with the Versions then, Ionic2 is like angular2, isn't that ok /?

guybedford:
@vadimpopa we don't currently support npm dist tags unfortunately. `jspm install npm:ionic2.0.0-alpha.9` would be the version to install here. (from `npm dist-tag ls ionic`). Support for dist-tags is tracking in https://github.com/jspm/npm/issues/61.

vadimpopa:
ok, got it, thanks.

vadimpopa:
@guybedford now with jspm install npm:ionic-framework is working, that's the correct ionic2 package

registry.json
```diff
@@ -234,6 +234,7 @@
   "infamous": "npm:infamous",
   "intro": "github:usablica/intro.js",
   "ionic": "github:driftyco/ionic-bower",
+  "ionic2": "ionic-framework",
```
guybedford: Did you mean `npm:ionic-framework`?

vadimpopa: right.

PR #635, opened 2015-10-27
Fix googlediff@0.1.0.json

mrydengren:
Apparently my fix in #634 exported an object literal and not the expected `diff_match_patch` function. This should fix that.

### DIFF
--- a/package-overrides/npm/googlediff@0.1.0.json
+++ b/package-overrides/npm/googlediff@0.1.0.json
@@ -1,4 +1,11 @@
 {
   "format": "global",
-  "main": "javascript/diff_match_patch_uncompressed.js"
+  "files": [
+    "javascript/diff_match_patch_uncompressed.js"
+  ],
+  "shim": {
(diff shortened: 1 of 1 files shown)


guybedford:
What was in the object literal? Was it definitely not the right object?

I'd suggest leaving out the `files` array and keeping the `main` though.

mrydengren:
[This](https://github.com/shimondoodkin/googlediff/blob/master/javascript/diff_match_patch_uncompressed.js#L2190-L2193) was in the object literal, but according to [`index.js`](https://github.com/shimondoodkin/googlediff/blob/master/index.js#L1) only the `diff_match_patch` function should be exported, which is expected by other modules using `googlediff`, such as [power-assert-formatter](https://github.com/power-assert-js/power-assert-formatter/blob/master/lib/udiff.js#L3).

guybedford:
Interesting, this would have worked equally well with just a shim on the `diff_match_path` file to make it be treated like a global then. But this is better to avoid the pointer main anyway.

PR #639, opened 2015-10-29
add an audio plugin for systemjs

ozsay:
this is an audio plugin for systemjs module loader

hopefully this is the right way to add packages to the registry.


registry.json
```diff
@@ -14,7 +14,8 @@
   "less": "github:aaike/jspm-less-plugin",
   "ts": "github:frankwallis/plugin-typescript",
   "sass": "github:screendriver/plugin-sass",
-
+  "audio": "github:ozsay/plugin-audio",
```
guybedford: Plugins go above in alphabetical order.

ozsay:
@guybedford I also fixed the order of all the plugins

guybedford:
:+1:

PR #647, opened 2015-11-09
repo renamed

capaj:
JSPM had probems creating public/jspm_packages/github/capaj/moonridge-client@0.9.1.js file, so I ditched the capital letter


guybedford:
Thanks, yes casing is currently delegated entirely to the registry implementations to deal with.

capaj:
@guybedford maybe we could throw error in the registry test if a repo name is with capital on the beginning? Is that a good idea?

guybedford:
Thanks, I've added https://github.com/jspm/jspm-cli/issues/1274 which would effectively make this unnecessary I think.

PR #648, opened 2015-11-09
reverse the order of the css files as dependencies

lookfirst:
@guybedford i'm not sure why this is suddenly failing the other way around, but doing it this way works. =)

### DIFF
--- a/package-overrides/github/mbenford/ngTagsInput@3.0.0.json
+++ b/package-overrides/github/mbenford/ngTagsInput@3.0.0.json
@@ -9,8 +9,8 @@
     "ng-tags-input.min": {
       "deps": [
         "angular", 
-        "./ng-tags-input.min.css!", 
-        "./ng-tags-input.bootstrap.min.css!"
+        "./ng-tags-input.bootstrap.min.css!",
+        "./ng-tags-input.min.css!"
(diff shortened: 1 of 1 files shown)


lookfirst:
@guybedford Actually... even more annoying... this behavior is different depending on whether or not the source code was built using the Builder or not... so don't merge this... the real issue here is that I think a dependency chain needs to be setup for these css files...

lookfirst:
http://mbenford.github.io/ngTagsInput/documentation/guides/bootstrap

lookfirst:
@guybedford tried this... no go... I'm probably doing something wrong here...

```
     "github:mbenford/ngTagsInput@3.0.0": {
        "main": "ng-tags-input.min",
        "registry": "jspm",
        "dependencies": {
          "angular": "^1.4.0",
          "css": "*"
        },
        "shim": {
          "ng-tags-input.min": {
            "deps": [
              "angular",
              "./ng-tags-input.min.css!",
              "./ng-tags-input.bootstrap.min.css!"
            ]
          },
          "ng-tags-input.bootstrap.min.css!": {
            "deps": [
              "./ng-tags-input.min.css!"
            ]
          }
        }
      }
 ```

PR #650, opened 2015-11-10
fix materialize override

jdanyow:


jdanyow:
the original override has `"files": ["dist"],` which results in empty directories

package-overrides/github/Dogfalo/materialize@0.97.0.json
```diff
@@ -1,10 +1,10 @@
 {
-  "files": [
-    "dist"
-  ],
-  "main": "dist/js/materialize",
+  "main": "js/materialize",
+  "directories": {
+    "lib": "dist"
```
guybedford: We avoid adding this property to overrides

guybedford:
Thanks for posting. Do you mean the directories apart from `dist` are empty, or is there an issue with directories being removed unnecessarily?

jdanyow:
the subdirectories were empty.  The install would create the directory structure but there would be no files within.  Maybe the `files` option doesn't do partial matches?  In any case, I think using the directories/lib option yields a better install result.

guybedford:
`files: ['dist']` will clear out all directories other than `dist`. If it was clearing anything inside `dist` that would be a bug though. `directories.lib` is a breaking change so we can't do that unfortunately.

jdanyow:
hmm- this came up because a user on stackoverflow hit the empty directory issue, which I reproduced when testing.  See comment [here](http://stackoverflow.com/questions/33614451/materialize-css-javascript-functions-not-found-in-aurelia-project/33632241?noredirect=1#comment55038054_33632241)

If you try the current override do you hit the same issue?  Maybe I'm doing something wrong.

guybedford:
The issue here is that it is using the release archive at https://github.com/Dogfalo/materialize/releases, which does not contain a `dist` folder.

So I think the correction here will work fine as you've posted without the `directories.lib` part.

PR #652, opened 2015-11-12
feat(AngularStompDK): add AngularStompDK 0.3.2 and 0.3.3

davinkevin:


guybedford:
Note that overrides apply for all versions semver compatible with the override, so this will apply for all `0.3.x`.

davinkevin:
Thanks ;)

PR #656, opened 2015-11-17
Added material, angular dep and removed css dependency (sass include)

DominicBoettger:


package-overrides/github/angular/bower-angular-aria@1.4.7.json
```diff
@@ -0,0 +1,12 @@
...
+  "shim": {
+    "angular-aria": {
+      "deps": ["angular"]
+    }
+  },
+  "registry": "jspm",
+  "dependencies": {
+    "angular": "^1.4.0"
+  }
+}
```
guybedford: This override shouldn't be necessary given https://github.com/jspm/registry/blob/master/package-overrides/github/angular/bower-angular-aria%401.3.0.json?

package-overrides/github/angular/bower-material@1.0.0-rc4.json
```diff
@@ -0,0 +1,16 @@
...
+    "angular-animate": "github:angular/bower-angular-animate@^1.4.7",
+    "angular-aria": "github:angular/bower-angular-aria@^1.4.7",
+    "css": "github:systemjs/plugin-css@^0.1.9"
+  },
+  "shim": {
+    "angular-material": {
+      "deps": [ "angular" ]
+    }
+  }
+}
```
guybedford: This override also shouldn't be necessary since it is in the original package.json file itself at - https://github.com/angular/bower-material/blob/master/package.json#L12.

DominicBoettger:
Alright, i will fix it in the original repository of material. Thanks.

PR #657, opened 2015-11-17
angular-aria version

DominicBoettger:
Latest angular-aria version

### DIFF
--- a/package-overrides/github/angular/bower-angular-aria@1.4.7.json
+++ b/package-overrides/github/angular/bower-angular-aria@1.4.7.json
@@ -1,0 +1,12 @@
+{
+  "main": "angular-aria",
+  "shim": {
+    "angular-aria": {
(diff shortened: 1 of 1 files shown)


package-overrides/github/angular/bower-angular-aria@1.4.7.json
```diff
@@ -0,0 +1,12 @@
...
+  "shim": {
+    "angular-aria": {
+      "deps": ["angular"]
+    }
+  },
+  "registry": "jspm",
+  "dependencies": {
+    "angular": "^1.4.7"
+  }
+}
```
guybedford: Again, this shouldn't be necessary given the existing override at https://github.com/jspm/registry/blob/master/package-overrides/github/angular/bower-angular-aria%401.3.0.json.

DominicBoettger:
okay. sorry!

PR #665, opened 2015-11-21
Override ReactJS to load the dist file

jackfranklin:
@guybedford this has been discussed in some prior issues about performance of jspm.

Particularly with React jspm has to load a lot of files, so when starting a project with jspm & React you start off with a slow time due to the number of modules that are imported when loading React.

This PR (if I've done the config correctly, which I have tested with `jspm install react -o ...`) tells jspm to load in the one React file instead.

However, @oliverjash pointed out that he likes React being loaded in bit by bit for nicer stack traces and easier debugging of a specific part of React. I would argue that most people don't need to dive into React, and the advanced use is still possible by doing `jspm install react -o "{}"` (I think? - might be nicer to add a `--no-override` flag or similar).

Additionally (and this might be a separate issue in SystemJS/Builder) it would be nice to be able to tell jspm to use a different file when bundling. The `dist/react.min.js` file is specially minified and has some dev code removed so it's more efficient and smaller than a bundle jspm will generate by default. If you could specify which file to use when bundling React that would be really a great help.

### DIFF
--- a/package-overrides/npm/react-@0.14.3.json
+++ b/package-overrides/npm/react-@0.14.3.json
@@ -1,0 +1,3 @@
+{
+  "main": "dist/react.js"
+}


guybedford:
I completely understand that the loading performance is an issue for separate files, but the issue with doing this is that it takes away the whole advantage of being able to share dependencies. All public dependencies of React then have to be duplicated if shared with other code, which goes against the principle of shared package management.

The way to handle this would be to create a bundle of React that excludes its public dependencies, and only does inlining of the internal dependencies. This sort of workflow is where `jspm build` is heading to make this easy for library authors to publish.

Since we don't have this for React, the real focus is to solve the root cause. That is to solve the issue of slow performance for many files in the browser. The way to do this is through pre-bundling on first load, and then hotreloading in the browser to match the webpack workflow, and jspm 0.17 is looking at the possibility of including these ideas.

jackfranklin:
@guybedford understood, thank you; that makes sense. I can still do this on a project by project basis if I decide to I guess :+1:

PR #670, opened 2015-11-24
add angular-toastr.

akash1810:
Adding [angular-toastr](https://github.com/Foxandxss/angular-toastr).


package-overrides/github/Foxandxss/angular-toastr@1.6.0.json
```diff
@@ -0,0 +1,17 @@
+{
+    "main": "angular-toastr.min",
```
guybedford: It's usually best to use the unminified versions for better debugging.

akash1810: ah, may bad. I've also not branched on my fork! Closing and will open a new PR.

Issue #659, opened 2015-11-17
lightbox2 is missing css and image files and the jquery dependency


sandroboehme:
When installing lightbox2 I only get the JavaScript file but not the CSS file and the image files. jQuery is also not automatically available after the installation.
I already forked lightbox2 to configure the package.json and install this version but with no success. Can someone please have a look?

Thanks in advance!

sandroboehme:
I've seen that there is an overwrite at https://github.com/jspm/registry/blob/master/package-overrides/github/lightbox2/lightbox2%402.7.1.json so I tried the version 2.7.1 (used 2.8.7 before). The override json file contains the jQuery dependency and the CSS file but no images. When reloading the page the CSS file and the images are sadly still not loaded.

sandroboehme:
I also tried 
```
./jspm install lightbox2 -o
"{
  "main": "./dist/js/lightbox.js",
  "dependencies": {
    "jquery": "^2"
  },
  "shim": {
    "js/lightbox": {
      "deps": [
        "jquery",
        "./dist/css/lightbox.css!",
    "./dist/images/close.png!",
    "./dist/images/loading.gif!",
    "./dist/images/next.png!",
    "./dist/images/prev.png!"
      ],
      "exports": "$"
    }
  }
}"
```
which displayed no error at the command line but still misses the files at the page.

guybedford:
If you look at the repo at https://github.com/lokesh/lightbox2, you can see that the files are in `dist/js/lightbox` not `js/lightbox`.

So the override would need to be:

```
  "main": "dist/js/lightbox.js",
  "dependencies": {
    "jquery": "*",
    "css": "*"
  },
  "shim": {
    "dist/js/lightbox": {
      "deps": [
        "jquery",
        "../css/lightbox.css!"
      ],
      "exports": "$"
    }
  }
```

If that works for you, please do send a PR to update the override here.

sandroboehme:
Thanks for the hint @guybedford! It looks like I did it right for the main property but missed the dist directory in the shim.
It works for me when I add the package overwrite to the package.json.
Now I'v added the content from the "overrides"."github:lokesh/lightbox2@2.8.1" property of the package.json to the lightbox2@2.8.1.json file in the registry and sent a PR with https://github.com/jspm/registry/pull/673. I hope it was the right thing to do.

PR #673, opened 2015-11-25
added package overwrite for lightbox2@2.8.1

sandroboehme:
Added the overwrite for the jQuery dependency, the missing css and the missing images. This also  fixes https://github.com/jspm/registry/issues/659
Thanks @guybedford for the hint regarding the package overwrite handling!


guybedford:
Looks great, thanks!

PR #675, opened 2015-11-27
Add angular dependency for angular-resource

ssljivic:


package-overrides/github/angular/bower-angular-resource@1.4.8.json
```diff
@@ -0,0 +1,13 @@
+{
+  "format": "cjs",
```
guybedford: Only global format works with shims currently, so it may be best to leave this out.

guybedford:
:+1:

PR #678, opened 2015-12-01
hls.js 0.2+ and 0.3+ overrides

dananichev:
Will it work for all 0.2.x and 0.3.x versions? 

Thanks!


guybedford:
The versions should be `0.2.0` and `0.3.0` and then it will apply to all `0.2.x` and `0.3.x`.

dananichev:
@guybedford done

PR #683, opened 2015-12-15
fix(chieffancypants.angular-hotkeys): add css to add information about hotkeys

davinkevin:


package-overrides/github/chieffancypants/angular-hotkeys@1.4.5.json
```diff
@@ -1,12 +1,18 @@
 {
-  "main": "build/hotkeys",
+  "directories": {
+    "lib": "build"
```
guybedford: This property will cause breaks for existing users so it is not encouraged.

guybedford:
:+1:

davinkevin:
I don't understand why it will break for existing users ? 
For that library, all the required files are under build (css or js), so we can override it.

If you prefer, I can modify my PR to use the previous syntax, but I would like to understand :smile: 

Thanks

guybedford:
Yes please do. It's because it alters the file / module name structure to no longer contain the `build` path, which will get pushed out to all existing users. Overrides already in use must be backwards compatible, at least until jspm 0.17.

davinkevin:
Modification done :smile:

PR #693, opened 2015-12-25
Add ember-wiz package

ricottatosta:


guybedford:
If you've got this info in your package.json at https://github.com/ricottatosta/ember-wiz/blob/master/package.json, then there's no need to post an override here though.

ricottatosta:
I would like users to be able to install the package just typyng as follow:
jspm install ember-wiz.
What am I supposed to do?

PR #704, opened 2016-01-03
Create lock@8.0.4

Zoltu:
Let me know if there is a better way to have an override apply to multiple versions.


Zoltu:
Getting it from bower appears to work out of the box, so this is unnecessary.  `jspm install bower:auth0-lock`

guybedford:
The `bower` registry in jspm is really not encouraged. Happy to merge this.

Zoltu:
Fixed broken build and re-opened PR.

package-overrides/github/auth0/lock@8.0.4.json
```diff
@@ -0,0 +1,9 @@
+{
+  "format": "global",
+  "main": "build/auth0-lock.min",
```
guybedford: Usually its best to use the non-minified version for debugging.

guybedford:
Did you try `npm:auth0-lock` here? Does that pull in too much cruft? It may be possible to remove the other stuff with an override similarly.

Zoltu:
`npm:auth0-lock` doesn't appear have the pre-built libraries (no build folder or similar).  I have switched to un-minified source.

Issue #709, opened 2016-01-08
Add package override for Bluebird to indicate it is a global module


drub0y:
Bluebird is a global module, but, because there's no package override for it and it comes from npm, JSPM's default logic is to treat it as a CommonJS module. This is a problem because one of the key scenarios for Bluebird right now is fixing [Edge's Promise performance problems](http://www.danyow.net/edge-promise-perf/) and that means including Bluebird _before_ SystemJS to establish the `Promise` type globally before SystemJS starts doing its work.

guybedford:
Can you clarify exactly what the issue is currently? Is it that jspm is introducing its own wrappers to the file so it can't be included with a script tag?

If so, then we must be careful to just treat that one specific file as a global and not the whole repo. This can be done with a shim override of the form:

```json
{
  "shim": {
    "path/to/bluebird-browser-without-js-extension": true
  }
}
```

Let me know if something along those lines can work for you here.

drub0y:
Hey @guybedford, sure let me explain. This is my first dip into the water with creating an override, so I appreciate any advice.

Bluebird is a polyfill for the Promise API. It is a single file only (at least in 3.1.1). The script is intended to be included via `<script>` tag and it adds or replaces the `window.Promise` global as well as the `window.P` global.

The specific use case is to actually deal with the fact that [Edge's Promise API implementation has some serious issues right now](http://www.danyow.net/edge-promise-perf/) that actually make SystemJS, which obviously makes heavy use of promises, perform terribly. So it needs to be included before SystemJS to replace that global Promise API before SystemJS starts doing its work.

guybedford:
@drub0y you are completely right for the browser use case, but jspm does also work in Node (especially with the 0.17 release coming out that works comprehensively in Node for universal js as well).

When you load Bluebird in NodeJS, it does not get treated like a global but as CommonJS.

Bluebird has two versions - one for the browser and one for Node, and this is distinguished by the browser field in its package.json.

Your solution is exactly correct, but we just need to do it for the browser file only.

**I was just asking if you could test out the following override for me:**

```
  jspm install npm:bluebird -o "{ shim: { "js/browser/bluebird": true } }"
```

and let me know if that works ok for you? If so lets get this fix into the registry.

drub0y:
@guybedford Ah, I understand 100% now. Chalk that up to y lack of experience, I was totally browser focused. Ok, I'll make the change and update the PR!

drub0y:
So two questions:

[...]

PR #710, opened 2016-01-08
Adding package override for bluebird 3.1.1

drub0y:
This addresses #709. I've tested locally with `jspm install -o` approach and it works smashingly.


guybedford:
Looks great, thanks for your work on this.

PR #712, opened 2016-01-14
Add can.js with override @2.3.9

Qazzian:
This tell jspm to use the prebuilt cjs modules for can.js
These are not available the github repo but are in the npm package.


Qazzian:
cool, thanks

PR #715, opened 2016-01-20
Adding override for npm:bootstrap-datepicker

mello151:


guybedford:
@mello151 this seems to work fine without the override from what I can tell?

mello151:
@guybedford sorry for the late reply. Yeah, it looks like it does. Sorry about that. I should've tested it without the override after switching over to the NPM package from the github one.

PR #718, opened 2016-01-21
Changed Source of Foundation

alfkonee:
zurb Foundation Source changed from bower to latest npm feed

### DIFF
--- a/package-overrides/npm/foundation-sites@5.5.1.json
+++ b/package-overrides/npm/foundation-sites@5.5.1.json
@@ -1,0 +1,14 @@
+{
+  "main": "js/foundation/foundation",
+  "format": "global",
+  "shim": {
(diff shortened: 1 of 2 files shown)


guybedford:
Thanks, does the npm version install and work correctly?

alfkonee:
Yes I've tested it with jspm install npm:foundation-sites and it works perfectly

guybedford:
There's a lot of stuff going on in https://github.com/jspm/registry/blob/master/package-overrides/github/zurb/bower-foundation%405.5.1.json, which I don't think would be captured?

alfkonee:
@guybedford Those overrides won't be captured but npm already resolves all of the dependencies and I updated the for npm:foundation-sites@5.5.1 without the ignores since they weren't needed

guybedford:
:+1: looks good.

PR #719, opened 2016-01-22
Update registry.json

JeroenVinke:
We renamed aurelia-kendoui-plugin to aurelia-kendoui-bridge and changed it to use npm


guybedford:
Sure, apologies for the delay here.

PR #726, opened 2016-01-31
Add videogular

whitneyit:
http://www.videogular.com/
https://github.com/videogular/bower-videogular


guybedford:
Is it not possible to use the npm version?

whitneyit:
It should be able to. I just defaulted to github our of habbit :smile:

guybedford:
Sure it's generally preferred and will come with things like `format: 'cjs'` by default.

whitneyit:
Do you want me to remove the `format` then?

whitneyit:
Also, will this download and install the required dependencies?

guybedford:
Yes it will need to be tested separately as there are different behaviours for npm packages.

whitneyit:
I ask because the [`package.json`](https://github.com/videogular/bower-videogular/blob/master/package.json) doesn't contain any reference to `angular` or `angular-sanitize`

guybedford:
Yes they would need to be added as dependencies.

PR #732, opened 2016-02-02
Create angular@1.5.0-rc.2.json

subesokun:
Lately I prefer installing Angular via NPM as this seems to be the recommended way (instead of using the bower github repo).

Is the usage here of `jspmNodeConversion` correct? The NPM package is not a jspm-style package but I hope the package overrides are fixing that. Should I also add the `"registry": "jspm"` property?

### DIFF
--- a/package-overrides/npm/angular@1.5.0-rc.2.json
+++ b/package-overrides/npm/angular@1.5.0-rc.2.json
@@ -2,10 +2,6 @@
   "format": "global",
   "jspmNodeConversion": false,
   "main": "angular",
-  "files": [
-    "angular.js",
-    "package.json"
-  ],
   "meta": {
     "angular.js": {
       "exports": "angular"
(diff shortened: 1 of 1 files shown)


package-overrides/npm/angular@1.5.0-rc.2.json
```diff
@@ -0,0 +1,14 @@
+{
+  "format": "global",
+  "jspmNodeConversion": false,
+  "main": "angular",
+  "files": [
+    "angular.js",
+    "package.json"
+  ],
```
guybedford: It's best not to include this, as otherwise it won't be obvious why the other files are missing.

guybedford:
Thanks, this looks great to me.

package-overrides/npm/angular@1.5.0-rc.2.json
subesokun: Ok, I'll remove the files filter again.

PR #733, opened 2016-02-02
rangy: use cjs instead of global

andrenarchy:
Fixes imports for rangy submodules, e.g. `import 'rangy/rangy-serializer'`;

### DIFF
--- a/package-overrides/github/timdown/rangy-release@1.2.3.json
+++ b/package-overrides/github/timdown/rangy-release@1.2.3.json
@@ -1,5 +1,5 @@
 {
-  "format": "global",
+  "format": "cjs",
   "main": "rangy-core",
   "shim": {
     "rangy-core": {


guybedford:
Can you clarify exactly what was breaking before and how this fixes it? Note that the modules already specified in the shim will still be loaded as globals as shim implies globals in jspm 0.16.

andrenarchy:
The code
```
import 'rangy';
import 'rangy/rangy-serializer';
```
failed in the second line because `rangy` is `undefined` [here](https://github.com/timdown/rangy-release/blob/master/rangy-serializer.js#L28). I noticed that neither AMD nor CommonJS are detected in the rangy modules and it uses `root.rangy` (which is undefined). Explicitly using `cjs` fixed it.

guybedford:
@andrenarchy the shims should exactly be making this work though. Which version of jspm are you using here?

andrenarchy:
0.17.0-beta.5

guybedford:
I just tried out these imports in jspm 0.17 and it is definitely working fine for me actually. Have you included the `jspm.browser.js` file before the `jspm.config.js` file in the test page?

andrenarchy:
Thanks for testing! You're right: it actually works if you just install and import rangy. After comparing our full config with the minimal example, I figured out that adding jquery makes it fail (!):
```
import 'jquery';
import 'rangy';
import 'rangy/rangy-serializer';
```
@guybedford Sounds like a bug in systemjs, right?

guybedford:
@andrenarchy it works fine for me... What do you mean by it makes it fail? Perhaps you have a circular reference?

andrenarchy:
I pushed my minimal example here: https://github.com/andrenarchy/jspm-tests

The browser throws the following error:
```
Error: Cannot read property 'createModule' of undefined(…)
```
Can you reproduce this?

guybedford:
Thanks, I managed to track this down and have posted the root case issue in https://github.com/systemjs/systemjs/issues/1075.

As a workaround try adding this configuration to jspm.browser.js:

```javascript
  meta: {
    "*": {
      scriptLoad: false
    }
  },
```

andrenarchy:
Thanks, this works! Do you have a pointer do documentation of the `scriptLoad` option or can you briefly explain what it does?

guybedford:
It's not actually documented, but it means SystemJS will use script tag injection over XHR + eval to load the module, making it CSP compatible as much as possible.

guybedford:
(effectively System-CSP is `scriptLoad: true` for all modules)

PR #736, opened 2016-02-03
bootstrap v4.0.0 alpha2

delebash:
add bootstrap v4.0.0 alpha2


guybedford:
Is the override already at https://github.com/jspm/registry/blob/master/package-overrides/github/twbs/bootstrap%404.0.0-alpha.json not applying for you here? Or is this specifically to include the tether dependency? Is that a dependency all users would want necessarily?

delebash:
it is for version2 of alpha 4.0.0-alpha-2. yes this includes tether, the first alpha did not use tether from what I can tell

Date: Wed, 3 Feb 2016 06:08:29 -0800

guybedford:
So this is just to add tether? Is it a required dependency?

delebash:
yes, and yes it is required

Date: Wed, 3 Feb 2016 06:48:17 -0800

guybedford:
:+1:

delebash:
thanks

Date: Wed, 3 Feb 2016 07:26:57 -0800

delebash:
I did something wrong here.  I tested it yesterday and it worked but I think I forgot that I had a link to tether in my html file so that is why it worked.

When using the current override Bootstrap does not see tether.js, trying to figure it out but at a loss right now

Date: Wed, 3 Feb 2016 07:26:53 -0800

guybedford:
It sounds like an override is needed for Tether then?

delebash:
thanks, that fixed it

Date: Wed, 3 Feb 2016 15:35:39 -0800

guybedford:
Could you share your override for Tether with another PR to the registry?

guybedford:
Ah just saw that now.

PR #745, opened 2016-02-07
feat(highcharts/highcharts-dist): add simpler support for highcharts

davinkevin:
@guybedford The repository of HighCharts add all modules to the same repository.
This overrides is the simpler version to support JSPM simply, but if you have a better solution, I'll be glad to hear it.

### DIFF
--- a/package-overrides/github/highcharts/highcharts-dist@4.2.2.json
+++ b/package-overrides/github/highcharts/highcharts-dist@4.2.2.json
@@ -1,0 +1,9 @@
+{
+  "main": "highcharts",
+  "shim": {
+    "highcharts": {
+      "deps": [ "./adapters/standalone-framework" ],
+      "exports": "Highcharts"
+    }
+  }
+}


guybedford:
Did you try `npm:highcharts`?

davinkevin:
This overrides is for another angular library (Highcharts-ng) which is only available on github and witch use HighCharts. The module seems to be written as a cjs module, but my override with npm doesnt' seem to work.

I've try to declare an override for this dependency (HighCharts-ng) like this 
```
    "overrides": {
      "github:pablojim/highcharts-ng@0.0.11": {
        "dependencies": {
          "highcharts": "npm:highcharts@^4.2.2",
          "angular": "github:angular/bower-angular@^1.2.0"
        },
        "shim": {
          "dist/highcharts-ng.js": {
            "deps": [
              "highcharts"
            ],
            "export": "Highcharts"
          }
        }
      }
    }
```
but when the code is executed, I have the following error
```
ReferenceError: Highcharts is not defined
    at initChart (http://localhost:8000/jspm_packages/github/pablojim/highcharts-ng@0.0.11/dist/highcharts-ng.js:357:23)
    at res.link (http://localhost:8000/jspm_packages/github/pablojim/highcharts-ng@0.0.11/dist/highcharts-ng.js:372:9)
    at http://localhost:8000/jspm_packages/github/angular/bower-angular@1.5.0/angular.js:9489:44
    at invokeLinkFn (http://localhost:8000/jspm_packages/github/angular/bower-angular@1.5.0/angular.js:9495:9)
    at nodeLinkFn (http://localhost:8000/jspm_packages/github/angular/bower-angular@1.5.0/angular.js:8981:11)
    at compositeLinkFn (http://localhost:8000/jspm_packages/github/angular/bower-angular@1.5.0/angular.js:8229:13)
    at compositeLinkFn (http://localhost:8000/jspm_packages/github/angular/bower-angular@1.5.0/angular.js:8232:13)
    at compositeLinkFn (http://localhost:8000/jspm_packages/github/angular/bower-angular@1.5.0/angular.js:8232:13)
    at publicLinkFn (http://localhost:8000/jspm_packages/github/angular/bower-angular@1.5.0/angular.js:8109:30)
    at link (http://localhost:8000/jspm_packages/github/angular/bower-angular-route@1.5.0/angular-route.js:1013:7) <div id="statsByType" config="sc.chartConfig" class="ng-isolate-scope">
```

Do you have an idea for that ? Do you think I need to open an issue instead of responding here ?

guybedford:
Yes this is a common issue because CommonJS definitions don't define globals (as they should).

The way around this is to use the `globals` metadata option, which allows a global to be set via meta just for that module:

```javascript
System.config({
  meta: {
    'path/to/highcharts-ng.js': { 
      'globals': {
        Highcharts: 'npm:highcharts@x.y.z'
      }
    }
  }
});
```

Ideally we should set this in the shim of the overrides, but that is only supported in jspm 0.17 not jspm 0.16 via `meta` configuration.

I'm tempted to get this 0.17 release out as soon as possible so we can no longer worry about having to revert to globals in situations like this. We can take PRs for the `meta` override in jspm 0.17 to make it work there if you want to upgrade.

Successful:
Do i still needs to write
` import * as Highcharts from 'highcharts';`

`instead of import 'highcharts'; `

because when i write the first one visual studio 2015 can't find the module, but it still works when running.

PR #750, opened 2016-02-09
angular-ui-select@0.14.0

### LINKED ISSUE #642: Are overrides version specific?
When I went to install angular-ui-select using `jspm install angular-ui-select` it pulled down version 0.13.2 which didn't seem to have the package overrides as specified [here](https://github.com/jspm/registry/blob/master/package-overrides/github/angular-ui/ui-select%400.9.6.json).

If I explicitly installed version 0.9.6 using `jspm install angular-ui-select@0.9.6`, it used the overrides (primarily including the CSS file).

Is it possible to make an override not specific to a version (or a range)?

techniq:
Add override for latest release of https://github.com/angular-ui/ui-select (0.14.2).  Closes #642

### DIFF
--- a/package-overrides/github/angular-ui/ui-select@0.14.2.json
+++ b/package-overrides/github/angular-ui/ui-select@0.14.2.json
@@ -1,0 +1,19 @@
+{
+  "dependencies": {
+    "css": "*",
+    "angular": "^1.3"
+  },
+  "directories": {
+    "lib": "dist"
+  },
+  "main": "select",
+  "registry": "jspm",
+  "shim": {
+    "select": {
+      "deps": [
+        "angular",
+        "./select.css!"
+      ]
+    }
+  }
+}
--- a/package-overrides/github/angular-ui/ui-select@0.14.2.json
+++ b/package-overrides/github/angular-ui/ui-select@0.14.2.json
@@ -3,16 +3,13 @@
     "css": "*",
     "angular": "^1.3"
   },
-  "directories": {
-    "lib": "dist"
-  },
-  "main": "select",
+  "main": "dist/select",
   "registry": "jspm",
   "shim": {
     "select": {
       "deps": [
         "angular",
-        "./select.css!"
+        "./dist/select.css!"
       ]
     }
   }


package-overrides/github/angular-ui/ui-select@0.14.2.json
```diff
@@ -0,0 +1,19 @@
+{
+  "dependencies": {
+    "css": "*",
+    "angular": "^1.3"
+  },
+  "directories": {
+    "lib": "dist"
```
guybedford: It's best to avoid adding this if you can as in case there are users using this already it will break for them.

guybedford:
:+1:

techniq:
I would send a PR to https://github.com/angular-ui/ui-select to update their package.json but since they currently have [183](https://github.com/angular-ui/ui-select/pulls) open, I figure it wasn't worth it.

I guess it would be helpful to have overrides for `0.10.x`, `0.11.x`, `0.12.x`, and `0.13.x`, which I guess requires a file for each?

guybedford:
It's fine to use an override and just focus on the latest version range, but `directories.lib` is a breaking change so even in the main package.json should be a major version bump.

techniq:
They recently tried to add support for npm/commonjs (see: https://github.com/angular-ui/ui-select/commit/79586aa4e32f4bc8f34cc9a5638341bda0930911, https://github.com/angular-ui/ui-select/commit/9f0e330e8303e83c1c31d8f164c1db2016f4a175) but `require('./dist/select.css');` causes havok and I couldn't figure out how to shim/etc a require for a css to use the css loader.

It would be nice if you could globally shim all require css statements to use the load.  I figure if it was `require('./dist/select.css!');` it would have used the loader...?

techniq:
I saw mention about directives.lib in the [docs](https://github.com/jspm/registry/wiki/Configuring-Packages-for-jspm#directorieslib) but wasn't sure what to do example (and based the override on 0.9.6 and other packages I saw.  Would this be better (and still work)?
```js
{
  "dependencies": {
    "css": "*",
    "angular": "^1.3"
  },
  "main": "dist/select",
  "registry": "jspm",
  "shim": {
    "select": {
      "deps": [
        "angular",
        "./dist/select.css!"
      ]
    }
  }
}
```

guybedford:
Yes exactly.

techniq:
Updated to not override `directories.lib`.

Btw, any thoughts on the `require('select.css')'.  It would have been nice to globally shim all "css require'd files" to use the css loader.

guybedford:
@techniq that is possible in jspm 0.17 via an override like:

```json
{
  "meta": {
    "*.css": {
      "loader": "css"
    }
  }
}
```

Overrides for jspm 0.17 are welcome in the registry too as long as they don't break 0.16 compatibility.

techniq:
I think I tried that but just noticed I'm running 0.16.19.  Guess I need to upgrade :D

techniq:
Oh, just noticed it's in beta :)

PR #753, opened 2016-02-11
Repoint d3 at npm.

mbostock:
With 4.0 coming, I’m moving generated files out of the D3 git repository. They’ll be available in npm (and a dedicated Bower git repository, but npm is preferred).

I tested this with `jspm install npm:d3` and it appeared to succeed?


guybedford:
Thanks for the PR, and for checking this support! I just tested it as well and it seems to work great. These registry entries are designed to be able to change without being restricted by backwards compatibility, so happy to land this.

mbostock:
Thank you!

PR #758, opened 2016-02-14
Add engine.io-client to registry

patrickheeney:


guybedford:
It may be better to use the `npm:engine.io-client` version here? Were there issues with running that?

patrickheeney:
I seem to have several issues with engine.io, or rather some module that is using it as a dependency. Related issue: https://github.com/jspm/jspm-cli/issues/1524 . Still trying to figure it out.

PR #759, opened 2016-02-14
Add webpack@1.12.13 to overrides

patrickheeney:


guybedford:
Thanks, this makes complete sense as jspm doesn't support arbitrary version ranges, so will always do a best-guess conversion (to ~0.6.0 in this case). Does Webpack even work in jspm 0.16 or did you test this in jspm 0.17?

patrickheeney:
I ran into these issues in latest 16, 17, and master. Related issue: https://github.com/jspm/jspm-cli/issues/1524

patrickheeney:
I only tested this fix in master though.

guybedford:
It's brave to be trying an install like this, although there's a chance it will work out in jspm 0.17 I guess! But knowing the right overrides to add is certainly not beginner territory.

PR #762, opened 2016-02-16
Adding scss plugin

KevCJones:
I've created a scss plugin based on mobilexags, however the use case was too different for his to be useful for me. Bare with me before you think its a cheap imitation, there is method to this madness.

The Sass plugin does a great job of cleanly injecting css from scss/sass into your head OR into a bundle (which in turn uses an IIFE to inject it into head). This is great, if you want that behaviour.  However, i didn't.

We're building an Angular2 seed internally (learning excercise / future planning stuff). Angular2's @component decorators allow you to pass 'styles' in as an array of style strings. Via this we are able to leverage it's shadow DOM emulation and management, lovely style encapsulation, with Angular removing stuff for us when that component is killed. 

So [using mobilexags as a base point] we've only slightly modified the injection arm to return the css to be used inline... and we've rebuilt the 'builder' arm a little more, to translate it into an 'amd' module for the bundle.

Using this then becomes a little more traditional (to me) module behaviour...

``` ts
import {Component} from 'angular2/core';
import styles from './app.css!';

@Component({
  selector: 'app',
  template: `
    <h1>{{ title }}</h1>
    <p>I will be styled using shadow-dom emulation - NG2 will manage DOM nicer than me</p>
  `,
  styles:[styles]
})

export class AppComponent {
    public title = 'WIP NG2 Seed Stuff - Angular 2 Boilerplate';
}
```

I found a great advantage was that i didn't need (with current tools out there) to use gulp or similar to build step the scss into css (using the text plugin to achieve the same) OR simply use the css file. 

I've modified his tests to reflect the changes in usage, later i'll add more tests that probably query the DOM instead of relying on acceptance-by-eye validating.

I figured, if this is useful for others you might want to add it to the registry. Now its out in the wild anyway if anyone wants to contribute, its there to be moulded for this particular use case. If we can find a way to achieve either behaviours with mobilexag's and this plugin then great, i'll happily fold this into his or visa versa.

Cheers

K

### DIFF
--- a/registry.json
+++ b/registry.json
@@ -16,6 +16,7 @@
   "plugin-typescript": "github:frankwallis/plugin-typescript",
   "raw": "github:matthewbauer/plugin-raw",
   "sass": "github:mobilexag/plugin-sass",
+  "scss": "github:KevCJones/plugin-scss",
   "text": "github:systemjs/plugin-text",
   "ts": "github:frankwallis/plugin-typescript",
   "tsv": "github:wpf500/plugin-tsv",


guybedford:
Some notes here:

* In https://github.com/KevCJones/plugin-scss/blob/master/src/index.js#L6 it looks like you're trying to do a build environment detection? If so the current API to do this is via `this.builder == true` in any hook itself.
* In https://github.com/KevCJones/plugin-scss/blob/master/src/index.js#L39 you can gain access to the shared pipeline data by just dumping any custom properties on `load.metadata`. Alternatively instead of implementing both fetch and translate, just implement fetch as returning an empty string, and put all the logic in translate rather.

* There is also a bundle hook which allows for cross-load handling and simplifications. In the CSS plugin this allows CSS to be minified across all CSS files, and dumped as a single blob of CSS in the build instead of separate CSS blobs for each module.

* This is my biggest worry with the approach here though - string-based per-module means that minification and shared style techniques are somewhat limited.

KevCJones:
Hey Guy,

Thanks for the feedback its appreciated.

Your final point i'll tackle first. Since i based this plugin on the already existing 'sass' one i suppose i've inherited some of the thinking behind it without to much consideration of improving it [yet], you are right though there are some things you need to be mindful of when creating the modules and tackling shared source, which is true i would say in all cases of module and source management. I'm not sure though why it would be limiting per say , but i'm sure you have a clear example you could fire my way and i'll see if i'd have some thought on how i'd get around it or build them in. I suppose it could be also how you use Sass that determines if you 'over import' source when really you could simply include 'shared' base styles as normal, i tended to see this approach friendlier if you were to import @mixins, and $variables to configure your components styles, but something like 'bootstrap-sass' for example would still be loaded and injected into head, giving me both approaches. If i'm missing a glaring oversight, i'm sure you'll take a little joy in pointing it out for me.  

Your other points are really helpful, its been a bit of 'learning on the job', when writing the workaround for passing data between hook, i knew there sure should be a way i could do this, but didn't see metadata. Makes perfect sense. 

I'm going to update, but i'd really appreciate the discussion to continue if you have time, i'm sure i'll find it educational to say the least.

K

KevCJones:
Updated with everything you said [ https://github.com/KevCJones/plugin-scss/blob/master/src/index.js]( https://github.com/KevCJones/plugin-scss/blob/master/src/index.js). Much cleaner!

Then i gave some thought to your bundle [hook] feedback and the cross loading and grouping of css.

There are indeed times when concat and bundling your css makes sense, global css's for example which would normally live in the head. 

I suppose the use-case i need and others who look at Angular2, where you want to leverage the [ViewEncapsulation.Emulated or Native](http://blog.thoughtram.io/angular/2015/06/29/shadow-dom-strategies-in-angular2.html), keeping some things separate makes sense then it pays to have those separate, but then it also pays to be able to leverage SCSS's @mixins and $vars too. 

Perhaps there is good argument for having an option to do both and being able to mark import's to use the 'bundle' hook but then have others have the option to be kept separated? Any thoughts on this?

Cheers

K

guybedford:
So that view section on Angular doesn't seem to provide information about how it handles CSS minification either. That is still my biggest concern with any type of style injection approach, but I guess that is a design decision here.

PR #763, opened 2016-02-19
Add chardin.js package

clintharris:


clintharris:
@guybedford Now that this is merged, I want to get my "manual" override out of my project's `package.json` file.

I ran the following thinking it would automatically pull down the newly-merged package overrides:
```
jspm uninstall heelhook/chardin.js
jspm cache-clear
jspm install github:heelhook/chardin.js
```
jspm doesn't seem to be pulling down the overrides though... Not seeing the new package in `~/.jspm/jspm-cache/package-overrides/github`.

Am I missing something?

clintharris:
@guybedford Cancel that--I see it now. Derp.

guybedford:
The other way to achieve this in jspm 0.17 is via `jspm install x -o` which will clear the override to the default.

PR #764, opened 2016-02-19
Add Materialize 0.97.5 package overrides

clintharris:


guybedford:
Is the override at https://github.com/jspm/registry/blob/master/package-overrides/github/Dogfalo/materialize%400.97.3.json not already working ok here?

clintharris:
@guybedford the json files are identical--however my project is "locked" in at v0.97.5, so I assumed we needed a new override file (even if it hasn't changed). Maybe I'm missing something?

When I run the following version-specific install (`jspm install github:Dogfalo/materialize@0.97.5`) I don't seem to be pulling down any overrides into `~/.jspm/jspm-cache/package-overrides/github/`.

guybedford:
@clintharris this is definitely working for me actually. Perhaps try clearing your cache, or uninstalling and then reinstalling?

clintharris:
Ah! You are correct--jspm is working fine and is getting the existing overrides.

That said, my manual override actually **isn't** the same as the existing ones and I think there's still a reason to consider the merge request.

Unlike the existing overrides, mine excludes  `"exports": "$"` because Materialize "exports" its own stuff (i.e., would otherwise put the following objects into the global namespace):
 - Materialize
 - Waves
 - Hammer 
 - etc.

One common example of why you need to get at these objects is to initialize the material "waves" effect:
```javascript
import materialize from 'Dogfalo/materialize';
...
materialize.Waves.displayEffect();
```
To me it makes more sense to not have the override export jQuery so that you can access the stuff it actually _does_ export.

Thoughts?

guybedford:
Sure, that makes sense. This PR doesn't have an `export` property set though?

clintharris:
What's the correct way to define an `exports` statement when the shimmed library adds more than one thing to the global namespace?

For example, this doesn't seem to work:
```
"exports": ["Materialize", "Waves", "Hammer", "Picker", "Vel", "validate_field"]
```
The only thing that gets exported is the first item in the list.

Thanks for any advice you can offer.

guybedford:
@clintharris feel free to +1 https://github.com/systemjs/systemjs/issues/443

clintharris:
Ah! Cool.

In the meantime, is it okay to leave out the `exports` statement from the override so that SystemJS can auto-detect all the globals and make them available?

guybedford:
Ah right I see what you mean. Yes, that's fine.

PR #766, opened 2016-02-20
Added override for Bootstrap 3.3.6

clintharris:
Provides overrides for Bootstrap 3.3.6.

Note that this differs from [bootstrap@3.3.4.json](https://github.com/jspm/registry/blob/master/package-overrides/github/twbs/bootstrap%403.3.4.json) in that it imports Bootstrap's CSS (as did [bootstrap@3.1.0](https://github.com/jspm/registry/tree/master/package-overrides/github/twbs)).

### DIFF
--- a/package-overrides/github/twbs/bootstrap@3.3.6.json
+++ b/package-overrides/github/twbs/bootstrap@3.3.6.json
@@ -6,7 +6,8 @@
       "deps": [
         "jquery",
         "../css/bootstrap.css!"
-      ]
+      ],
+      "exports": "$"
     }
   },
   "dependencies": {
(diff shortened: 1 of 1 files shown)


guybedford:
We disabled the CSS on Bootstrap due to user requests who had manual CSS workflows.

The goal was to re-enable it when we had a workflow for optionally adding the CSS.

That workflow is now possible in jspm 0.17 through the SystemJS alias feature via an override config like:

```json
{
  "main": "dummy-main-with-css"
  "meta": {
    "dummy-main-with-css": {
      "alias": "./js/bootstrap",
      "deps": ["./css/bootstrap.css"]
    }
  }
}
```

But the remaining issue with this approach is that we do not yet provide build support for alias modules (tracking in https://github.com/systemjs/builder/issues/422).

guybedford:
(in the above config `System.import('bootstrap')` would come with CSS, while `System.import('bootstrap/js/bootstrap.js')` would allow opting out of CSS.

clintharris:
Understood--thanks for the detailed explanation.

PR #768, opened 2016-02-21
Create bower-angular-messages@1.5.0.json

Offirmo:
Shim for angular-messages https://github.com/angular/bower-angular-messages

Exactly the same as https://github.com/jspm/registry/blob/master/package-overrides/github/angular/bower-angular-animate%401.3.0.json or https://github.com/jspm/registry/blob/master/package-overrides/github/angular/bower-angular-aria%401.3.0.json

This is my 1st registry pull request, please take care of me ! :bow:


guybedford:
:+1:

PR #777, opened 2016-02-24
Add highcharts

nicolashemonic:
https://github.com/highcharts/highcharts


package-overrides/github/highcharts/highcharts@4.2.3.json
```diff
@@ -0,0 +1,6 @@
+{
+    "main": "highcharts.src",
+    "directories": {
+        "dist": "lib"
```
guybedford: It's advisable not to add this as it will be a breaking change for any existing jspm users of this package.

PR #783, opened 2016-02-25
Create svgo@0.6.1.json

MeoMix:
Added deps to support config.js dynamic dependency loading.


MeoMix:
Is this something that could be added to svgo's package.json? Or is it cool staying in the registry?

guybedford:
Yes it's always better to check these into the projects' package.json.

MeoMix:
Okay. I'll make an attempt at it tonight when I'm done contracting.

PR #784, opened 2016-02-26
Delete engine.io-client@1.5.1.json

salfield:
This is causing an old and broken version of has-cors to get installed when you try and install socket.io.

Without it I am able to install socket.io, pending a proper fix to:

jspm/jspm-cli#1571


package-overrides/npm/engine.io-client@1.5.1.json
```diff
@@ -1,15 +0,0 @@
...
-      "component-emitter": "1.1.2",
-      "indexof": "0.0.1",
-      "engine.io-parser": "1.2.1",
-      "debug": "2.1.3",
-      "parseuri": "0.0.4",
-      "parsejson": "0.0.1",
-      "parseqs": "0.0.2",
-      "component-inherit": "0.0.3"
-  }
-}
```
guybedford: Rather than deleting, we should create a new override for the later version that this isn't needed on.

salfield:
Sure, that makes sense. However, the has-cors dependency can't actually be installed at 1.0.3 due to the use of a git archive as a dependency - so I can't see how this override will work in any case:

component/has-cors#1
component/has-cors#4

salfield:
This now contains everything needed to get socket.io installed and running. I've re-added the (imo broken) version of engine.io-client and added an empty one which allows the latest version to install. I'm not clear why a requirement for engine.io@1.6.8 would be affected by an override for 1.4.5 in any case?

Socket.io still cannot run using the npm version due to three issues:
1. jspm/jspm-cli#1571 - temporarily fixed without any obvious adverse effects in my fork of jspm
2. socketio/socket.io#2463
3. jspm/npm#88

Issues 2&3 are fixed in: salfield/socket.io

guybedford:
Can you remind me if you're running jspm 0.16 or jspm 0.17 here?

salfield:
0.17.0-beta.9

Issue #795, opened 2016-03-04
Angular2 Support


guybedford:
When installing `jspm install npm:angular2` unnecessary browserify deps get included.

This issue can track the creation of the necessary override for Angular, before check-in to the original package.json.

amcdnl:
Here is the override fix:

```
{
  "dependencies": {
    "zone.js": "npm:zone.js@0.6.2",
    "rxjs": "npm:rxjs@5.0.0-beta.2",
    "reflect-metadata": "npm:reflect-metadata@0.1.3"
  },
  "ignore": [
    "examples"
  ],
  "map":{
    "crypto": "@empty"
  },
  "jspmNodeConversion": false
}
```

guybedford:
Created https://github.com/jspm/registry/pull/797.

amcdnl:
PR'd https://github.com/jspm/registry/pull/798

amcdnl:
LOL u beat me!

architruc:
It seems that the [latest commit](https://github.com/jspm/registry/commit/6449bde5222d55723c93dc72e6b05a15b7c54bc0) on `angular2@2.0.0-beta.8.json` broke the JSPM bundling (thus systemjs-builder I guess).

This was added:

```json
"meta": {
  "src/core/util/decorators.js": {
    "deps": [
      "reflect-metadata"
    ]
  }
}
```

I am not sure (I am not familiar at all with this part of JSPM), but I guess it updated this file fetched by JSPM when installing Angular 2: `jspm_packages/npm/angular2@2.0.0-beta.12.json`

Since then, bundling with JSPM (0.17 beta 11) then loading on the browser results in the following error for me:

```
Cannot resolve all parameters for 'ResolvedMetadataCache'(?, ?). Make sure that all the parameters are decorated with Inject or have valid type annotations and that 'ResolvedMetadataCache' is decorated with Injectable.
```

```
angular2-polyfills.js:332 Error: TypeError: Cannot read property 'getOptional' of undefined(…)
```

The error has gone after I updated `jspm_packages/npm/angular2@2.0.0-beta.12.json` and replaced its content by an empty hash instead:

```json
{}
```

Any impact on this change? Would this issue be fixed?

Thanks a lot!

architruc:
I case someone else reads my previous post, the issue seems to have gone with more recent versions of Angular 2 (Checked on RC3). The above workaround is not needed any more.

PR #797, opened 2016-03-04
Angular2 override

guybedford:
Override for https://github.com/jspm/registry/issues/795.

//cc @amcdnl

It would be good to get confirmation from others using Angular if this is all good. @robwormald perhaps you could take a look here or suggest someone who could?

### DIFF
--- a/package-overrides/npm/angular2@2.0.0-beta.8.json
+++ b/package-overrides/npm/angular2@2.0.0-beta.8.json
@@ -10,5 +10,10 @@
     "rxjs": "npm:rxjs@^5.0.0-beta.2",
     "zone.js": "npm:zone.js@^0.5.15"
   },
-  "jspmNodeConversion": false
+  "jspmNodeConversion": false,
+  "meta": {
+    "src/core/util/decorators.js": {
+      "deps": ["reflect-metadata"]
+    }
+  }
 }
--- a/package-overrides/npm/angular2@2.0.0-beta.8.json
+++ b/package-overrides/npm/angular2@2.0.0-beta.8.json
@@ -1,14 +1,14 @@
 {
+  "main": false,
   "dependencies": {
-    "zone.js": "npm:zone.js@0.6.2",
-    "rxjs": "npm:rxjs@5.0.0-beta.2",
-    "reflect-metadata": "npm:reflect-metadata@0.1.3"
+    "reflect-metadata": "npm:reflect-metadata@^0.1.3",
+    "rxjs": "npm:rxjs@^5.0.0-beta.2",
+    "zone.js": "npm:zone.js@^0.5.15"
   },
-  "ignore": [
-    "examples"
-  ],
-  "map":{
-    "crypto": "@empty"
+  "peerDependencies": {
+    "reflect-metadata": "npm:reflect-metadata@^0.1.3",
+    "rxjs": "npm:rxjs@^5.0.0-beta.2",
+    "zone.js": "npm:zone.js@^0.5.15"
   },
   "jspmNodeConversion": false
 }
--- a/package-overrides/npm/angular2@2.0.0-beta.8.json
+++ b/package-overrides/npm/angular2@2.0.0-beta.8.json
@@ -1,0 +1,14 @@
+{
+  "dependencies": {
+    "zone.js": "npm:zone.js@0.6.2",
+    "rxjs": "npm:rxjs@5.0.0-beta.2",
+    "reflect-metadata": "npm:reflect-metadata@0.1.3"
+  },
+  "ignore": [
+    "examples"
+  ],
+  "map":{
+    "crypto": "@empty"
+  },
+  "jspmNodeConversion": false
+}
--- a/package-overrides/npm/reflect-metadata@0.1.3.json
+++ b/package-overrides/npm/reflect-metadata@0.1.3.json
@@ -1,0 +1,3 @@
+{
+  "jspmNodeConversion": false
+}
--- a/package-overrides/npm/zone.js@0.5.15.json
+++ b/package-overrides/npm/zone.js@0.5.15.json
@@ -1,0 +1,3 @@
+{
+  "dependencies": {}
+}
--- a/package-overrides/npm/zone.js@0.5.15.json
+++ b/package-overrides/npm/zone.js@0.5.15.json
@@ -1,3 +1,10 @@
 {
+  "main": "dist/zone.js",
+  "jspmNodeConversion":false,
+  "shim": {
+    "dist/zone": {
+      "exports": ["Zone", "zone"]
+    }
+  },
   "dependencies": {}
 }
--- a/package-overrides/npm/zone.js@0.5.15.json
+++ b/package-overrides/npm/zone.js@0.5.15.json
@@ -1,5 +1,6 @@
 {
   "main": "dist/zone.js",
+  "directories": {},
   "jspmNodeConversion":false,
   "shim": {
     "dist/zone": {


amcdnl:
If it helps, you can see the overrides in action w/ this repo - https://github.com/amcdnl/angular2-upgrading

amcdnl:
Can we go ahead and merge? I know @robwormald is very busy and this would really help me out a lot locally.

guybedford:
This is a change that will need to remain backwards compatible, and should be considered part of the Angular API - I'd only want to merge with confirmation from someone on the Angular team.

amcdnl:
Alright I understand, anyone else we can hit up to review?

guybedford:
@jeffwhelpley this is quite an important registry entry to get Angular2 correctly configured for jspm. Do you know anyone from Angular who may be able to review this PR before we go ahead here?

jeffwhelpley:
@guybedford yes, the best ones to look at this would be @gdi2290 or @jeffbcross. I will ping them on slack as well to make sure they see this and check it out.

guybedford:
That would be amazing, thanks.

package-overrides/npm/angular2@2.0.0-beta.8.json
```diff
@@ -0,0 +1,14 @@
+{
+  "dependencies": {
+    "zone.js": "npm:zone.js@0.6.2",
```
jeffbcross: I believe beta.8 depends on zone.js@0.5.15

gdi2290:
The current version if Angular 2 beta 8 is using zone.js 0.5.15

package-overrides/npm/angular2@2.0.0-beta.8.json
jeffbcross: And why are es6-promise and es6-shim not included? See package.json: https://github.com/angular/angular/blob/master/modules/angular2/package.json

amcdnl: @jeffbcross are those really required? I'm running babel which has these polyfilled already.

amcdnl: Its not a big deal to just leave these out altogether, all we really need is:
https://github.com/jspm/registry/blob/angular/package-overrides/npm/angular2%402.0.0-beta.8.json#L13 and the `empty` flag

robwormald:
sorry, didn't see this earlier. I believe JSPM should remain consistent with the NPM peer deps, so really the override should be similar to https://github.com/angular/angular/blob/master/modules/angular2/package.json#L12-L16

I'm happy to discuss eventually trimming that, but let's stick to what we know works, for the moment - otherwise i'll end up with 5000 tweets about it when it inevitably breaks... Most users are using TS, so we can't rely on the babel coverage. Babel uses can certainly choose not to import them if they don't need them, but I don't think there's any harm in the extra install there (as opposed to the original dozen+ extraneous deps, which is why the original issue got opened up on our end)

amcdnl:
@robwormald thats totally fine. The big thing we REALLY need is:

```
"map":{
    "crypto": "@empty"
  },
  "jspmNodeConversion": false
```

@guybedford : lets ditch the rest and just do the above if thats ok?

robwormald:
where is crypto coming from? reflect-metadata? maybe that's where the map/metaconfig should be done then?

amcdnl:
@robwormald cyrpto comes from angular when its running in Node mode.

guybedford:
Thanks for all the feedback here, much appreciated for the help to make sure this goes smoothly, as these overrides do form part of the "breakable" API.

* I was able to remove the `crypto` mapping since we have `jspmNodeConversion: false` set jspm will not attempt to automatically install crypto. The only issue will be that modules that require crypto will fail. Looking through the codebase this was in bundles/angular2-polyfills.js and ts/typings/node/node.d.ts. @amcdnl I take it you were loading the angular2-polyfills.js file then that you needed this, which was then giving a loading error? In that case, it may be better to set a `meta: { "bundles/angular2-polyfills.js": { cjsRequireDetection: false } }` to disable static analysis, but still allow crypto to resolve if it does exist in the registry. But that feature won't work in jspm 0.16 unfortunately. Otherwise I can add back the empty map too.
* I've set the dependencies as both dependencies and peerDependencies, so that in jspm 0.16 the dependencies will install as normal dependencies, and in jspm 0.17 the dependencies will install as peerDependencies.
* I've also set the dependencies to semver ranges. While this specific beta release is locked, the next beta will bump the ranges, and this override will continue to apply. That could potentially cause future releases to break as this override wouldn't update with them. Because jspm is locked by default, it shouldn't be as much of a problem to allow semver ranges in the package.json here.

amcdnl:
@guybedford -

> The only issue will be that modules that require crypto will fail. Looking through the codebase this was in bundles/angular2-polyfills.js and ts/typings/node/node.d.ts. @amcdnl I take it you were loading the angular2-polyfills.js file then that you needed this, which was then giving a loading error

Yes, polyfills are required to run Angular2 so when I include them it brings in crypto.

> In that case, it may be better to set a meta: { "bundles/angular2-polyfills.js": { cjsRequireDetection: false } } to disable static analysis, but still allow crypto to resolve if it does exist in the registry. But that feature won't work in jspm 0.16 unfortunately. Otherwise I can add back the empty map too.

I'm not running 17 yet, I'm waiting for it to RC before hopping on. We have our bundler builders and whatnot that will need to be upgraded and wanted things to stabilize out before I move.

Here is a small demo of A2 running with JSPM = https://github.com/amcdnl/angular2-upgrading

robwormald:
@amcdnl generally, you shouldn't be importing the polyfills file anyway - zone needs to do its thing before the module loader boots, or you can get caught in all kinds of tricky zone issues.

guybedford:
@amcdnl let me know if you can work with Rob's suggestion here?

amcdnl:
@robwormald hmmm ... that might explain some of my change detection issues!
@guybedford Yes, I'm good. I mainly only wanted the `jspmNodeConversion:false` and `crypto:empty`.

Lets call it done! :)

guybedford:
Please let's not call it done without 100% certainty. This is like publishing any release - the API should be carefully considered to be sure we can maintain backwards compatibility.

Specifically what still worries me is if the angular2-polyfills.js file shouldn't be loaded manually, what will that affect here specifically? What usually loads this file?

amcdnl:
@guybedford no problem! So @robwormald correct me if I'm wrong but if you don't load it in the right order the change detection won't trigger correctly. Angular2 hooks into events of the browser and without loaded in the correct order, angular core can't update correctly.

guybedford:
I've gone through this again against a test require of `System.import('angular2/core.js')` and `System.import('angular2/platform/browser.js')` in both jspm 0.16 and jspm 0.17 in both the browser and Node.

I was able to make the further simplifications:
* Add an override for zone.js to use the single-file build over the multifile unlinked version, this is for faster loading
* Update the reflect-metadata override to pull in fewer dependencies as it doesn't need Node shims
* In the jspm 0.17 case, automatically load the `reflect-metadata` shim when loading `src/core/util/decorators.js` in Angular. I really don't understand why this isn't already pulled in automatically?

My experience using Angular is very limited, so it would help to get feedback on further edge cases here... what other requires should be checked here beyond the above? Is `angular-polyfills.js` loaded manually by users, or is it not necessary to load at all?

There's likely a lot more that could be done here to improve this loading experience. Things like checking code branching between Node and browser (there doesn't seem to be any branching at the module level), using builds of private modules like I've done with the zone.js override to avoid many separate files (rxjs seems suitable for the separate files though as it is so heavily modular, but maybe more could be done here in Angular itself?). Always happy to discuss this stuff further.

Another thing worth considering is whether we should use the TypeScript source for angular by default, and then auto-configure the TypeScript plugin to apply correctly with Type checking in the Angular config. That might be quite a nice workflow too, depending on performance.

As soon as this is merged, I'll also post PRs to all the angular repos to have this config in the original package.json files too, although no pressure for merging those.

amcdnl:
@guybedford - awesome notes! 

Note: We should probably bump this to beta9 now since thats the latest. There are no major changes between versions, bug fixes and router additions.

guybedford:
Ok, I'm merging this. If there are any angry users, just let me know :)

guybedford:
~~Angry~~ user number 1 - https://github.com/jspm/jspm-cli/issues/1651.

guybedford:
I've created a correction PR at https://github.com/jspm/registry/pull/846.

amcdnl:
:( - we've been running and not had this issue.

PR #798, opened 2016-03-04
angular2 overrides

amcdnl:


aluanhaddad:
The referenced package is an old beta from before the angular team reconstructed their entire repository. Current releases are supported so I am going to close this.

PR #799, opened 2016-03-04
Create postcss-import@8.0.2.json

MeoMix:
See here: https://github.com/postcss/postcss-import/issues/181

`load-content` errors out when attempting to run in the browser. This isn't a big deal as it's module is overridden with options passed into PostCSS-Import. However, the module needs to not error out. So, I add an override for the browser environment which ensures an empty response.

QUESTION:

It's not clear to me whether I need to specify all other mapped dependencies in this file, or if I should only specify the one property.

Also, it's not clear to me how I would go about submitting a PR for something like this to postcss-import. I know this registry is supposed to be a temp location, but this seems specific to JSPM and not to their plugin.

### DIFF
--- a/package-overrides/npm/postcss-import@8.0.2.json
+++ b/package-overrides/npm/postcss-import@8.0.2.json
@@ -1,0 +1,10 @@
+"map": {
+  "object-assign": "npm:object-assign@4.0.1",
+  "postcss": "npm:postcss@5.0.19",
+  "postcss-value-parser": "npm:postcss-value-parser@3.3.0",
+  "read-cache": "npm:read-cache@1.0.0",
+  "resolve": "npm:resolve@1.1.7",
+  "./lib/load-content": {
+    "browser": "@empty"
+  }
+}


package-overrides/npm/postcss-import@8.0.2.json
```diff
@@ -0,0 +1,10 @@
+"map": {
```
guybedford: This needs to be a JSON file

package-overrides/npm/postcss-import@8.0.2.json
```diff
@@ -0,0 +1,10 @@
...
+  "object-assign": "npm:object-assign@4.0.1",
```
guybedford: Maps shouldn't include exact versions because version dependency ranges are used in the package dependencies.

package-overrides/npm/postcss-import@8.0.2.json
```diff
@@ -0,0 +1,10 @@
...
+  "postcss": "npm:postcss@5.0.19",
+  "postcss-value-parser": "npm:postcss-value-parser@3.3.0",
+  "read-cache": "npm:read-cache@1.0.0",
+  "resolve": "npm:resolve@1.1.7",
+  "./lib/load-content": {
+    "browser": "@empty"
```
guybedford: This object map is only supported in jspm 0.17. Rather use a `browser` field like in npm.

guybedford:
I think you can remove the map entirely and just use a Browserify `browser` field here rather.

For sending to the original repo, such a config is jspm-specific yes, but can be sent via `"jspm": { "browser": {  "x": "..." } }` in the package.json to be jspm specific in the original repo.

MeoMix:
Okay, I've submitted a PR to postcss-import w/ the package change.

PR #808, opened 2016-03-10
Added AdminLTE@2.3.2

LazerFX:


guybedford:
You may want to provide a shim config here as well for the jQuery dependency like:

```json
{
  "shim": {
    "dist/js/app": {
      deps: ["jquery"]
    }
  }
}
```

LazerFX:
Thanks for the suggestion - First time doing this, and I've not used the JSPM override system very much - that's actually made things better in my local build as well :+1:

guybedford:
Great, thanks for sending the PR here.

PR #813, opened 2016-03-12
Add functify

nlwillia:


package-overrides/npm/functify@0.4.0.json
```diff
@@ -0,0 +1,3 @@
+{
+  "main": "src/functify"
```
guybedford: I think for this ES main to work in jspm 0.16 might require `jspmNodeConversion: false` as well here? Let me know.

nlwillia:
I was not familiar with that option.  It doesn't seem to make a difference in use, but it looks like it may apply since this is not using the default published npm format but going after the ES6 source directly.  The reason I found that necessary is because the dist version references `regeneratorRuntime` which is a babel polyfill that I'm not sure how to otherwise incorporate.  Just letting the whole thing transpile on the fly in the usage environment produces a working result.

guybedford:
Sure, thanks for updating. Regenerator runtime is possible to shim through a dependency on that library, but we do need to start allowing workflows that publish ES6. There aren't that many overrides that do full ES6 yet, so it will be somewhat experimental here too to see how it works out.

PR #815, opened 2016-03-13
Add override for CryptoJS

dinoboff:
Might be useful over crypto when converting legacy code to JSPM.

### DIFF
--- a/package-overrides/github/sytelus/cryptojs@3.1.2.json
+++ b/package-overrides/github/sytelus/cryptojs@3.1.2.json
@@ -1,0 +1,221 @@
+{
+  "main": "core.js",
+  "format": "global",
+  "directories": {
+    "lib": "components"
+  },
+  "shim": {
+    "aes": {
+      "deps": [
+        "cryptojs/enc-base64.js",
+        "cryptojs/md5.js",
+        "cryptojs/evpkdf.js",
+        "cryptojs/cipher-core.js"
+      ],
+      "exports": "CryptoJS.lib.BlockCipher.AES"
+    },
+    "cipher-core": {
+      "deps": [
+        "cryptojs/core.js"
+      ],
+      "exports": "CryptoJS"
+    },
+    "core": {
+      "exports": "CryptoJS"
+    },
+    "enc-base64": {
+      "deps": [
+        "cryptojs/core.js"
+      ],
+      "exports": "CryptoJS.enc.Base64"
+    },
+    "enc-utf16": {
+      "deps": [
+        "cryptojs/core.js"
+      ],
+      "exports": "CryptoJS.enc.Utf16LE"
+    },
+    "evpkdf": {
+      "deps": [
+        "cryptojs/core.js"
+      ],
+      "exports": "CryptoJS.algo.EvpKDF"
+    },
+    "format-hex": {
+      "deps": [
+        "cryptojs/core.js"
+      ],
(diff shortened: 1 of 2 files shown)


package-overrides/github/sytelus/cryptojs@3.1.2.json
```diff
@@ -0,0 +1,129 @@
...
+    "lib": "components"
+  },
+  "shim": {
+    "core.js": {
+      "exports": "CryptoJS"
+    },
+    "sha224.js": {
+      "deps": [
+        "cryptojs/sha256.js"
+      ]
```
guybedford: I'm wondering if it's worth setting exports here as well - `"exports": "CryptoJS.algo.SHA224"` sort of thing?

guybedford:
Looks great!

package-overrides/github/sytelus/cryptojs@3.1.2.json
dinoboff: It uses `CryptoJS` global reference to add a SHA224 item to `CryptoJS.algo`. It doesn't export `CryptoJS.algo.SHA224`; it doesn't even export `CryptoJS` if it's not present.

Requiring `core.js` (via `"*": {"deps": ["cryptojs/core.js"]}`) which exports `CryptoJS` is not enough for SystemJS?

guybedford: Certainly it would be enough to make it work but it might be quite nice to
get back the direct module value for sha224 instead of the same global
object for each independent import.

dinoboff: I see.

It often extend CryptoJS and Crypto.algo; e.g. `sha512.js` adds `CryptoJS.SHA512` and `CryptoJS.algo.SHA512`. The former is a helper function to create the hash directly, the later is an object that other Crypto component will need (eg., you would pass it the pbkdf2 function to define which hasher it should use).

Would it work with something that?
```
    "sha512.js": {
      "deps": [
        "cryptojs/x64-core.js"
      ]
      "exports": {
         "hasher": "CryptoJS.SHA512",
         "algo": "CryptoJS.algo.SHA512"
      }
    },
```

Then could it be imported like that?
```
import {hasher, algo} from 'cryptojs/sha512.js';
```

guybedford:
Right but as an end-user surely you only want the `CryptoJS.SHA512` value? There isn't currently the ability to define exports as anything other than an array or a string.

dinoboff:
Updated:

- I removed the .js extension from shim key (It wouldn't shim the the components otherwise)
- I removed invalid entries (like hmac-md5 which is not a component but a rollup)
- added exports for every entries:
  * the only item the component adds to CryptoJS;
  * if it adds many and one is an helper function, it exports the helper function;
  * if it adds more than one helper function (a hasher and a hmac helper), it exports the hasher;
  * otherwise it exports CryptoJS.

dinoboff:
Is it possible to create an alias like "cryptojs/hmac-md5 -> md5" that exports the hmac helper function instead of the hasher?

guybedford:
You can do aliases, but it must be an alias to another module, not another shim of the same module.

Perhaps just do `import CryptoJS from 'cryptojs'; import 'cryptojs/md5'; CryptoJS.hmac-md5` or whatever the correct naming is for that case.

guybedford:
:+1:

PR #817, opened 2016-03-15
Update angularfire override with version 1.1.4

dinoboff:


guybedford:
Thanks! Unfortunately because we missed the major release here already long ago, this will be a breaking change for anyone using it already with custom shims. We should remove the directories.lib override to ensure it is non-breaking.

dinoboff:
@guybedford removed

package-overrides/github/firebase/angularfire@1.1.4.json
```diff
@@ -0,0 +1,16 @@
+{
+  "main": "dist/angularfire.js",
+  "shim": {
+    "*": {
```
guybedford: This should just be `"dist/angularfire"` I think?

guybedford:
Thanks please let me know when you've updated again re the comment above.

dinoboff:
@guybedford updated

guybedford:
:+1:

PR #818, opened 2016-03-15
Adding NPM Angular 1.5 + NPM Angular Route 1.5

barefootdeveloper:


guybedford:
The decision was originally made to use the Angular1 version from github, and that is what is provided by `jspm install angular`. Is there a reason you can't use this version?

barefootdeveloper:
Oh probably not. Just used to pulling from NPM and I saw angular 1.5rc2 out there so figured I might as well the release version. 

Should we delete 1.5rc2 then as well?

guybedford:
Sorry I forgot we had that. Actually the override at https://github.com/jspm/registry/blob/master/package-overrides/npm/angular%401.5.0-rc.2.json should already be applying for the 1.5.0 stable version as `^1.5.0-rc.2` includes `1.5.0`.

So there's no harm merging this, or we can keep it simple and just merge the angular route override.

barefootdeveloper:
It's up to you, as you indicated I switched to using github and it worked perfectly fine.  If you'd like to include it I'm happy to make any changes you request.  If you'd rather just close it and stick with using github that works for me too.

Thanks for the work you do with this, it works really slick :)

guybedford:
Sure, thanks, let's stick with the GitHub versions for now then unless this is necessary to come back to.

PR #819, opened 2016-03-15
npm:element-resize-detector override

MeirionHughes:


package-overrides/npm/element-resize-detector@1.1.0.json
```diff
@@ -0,0 +1,9 @@
+{
+  "main": "dist/element-resize-detector.min",
```
guybedford: It's best to use the unminified source on jspm for debugging purposes.

MeirionHughes:
Okay, updated it.

package-overrides/npm/element-resize-detector@1.1.0.json
```diff
@@ -0,0 +1,9 @@
+{
+  "main": "dist/element-resize-detector",
+  "format": "cjs",
+  "shim": {
+    "dist/element-resize-detector": {
+      "exports": "utils"
+    }
+  }
```
guybedford: Is this shim really necessary here? Does it not work ok without it?

guybedford:
Thanks, just noticed it should in theory work as CommonJS through the dist file though?

package-overrides/npm/element-resize-detector@1.1.0.json
MeirionHughes: yeah, during bundling it says it cannot resolve that utils, all the other exports are fine though.

MeirionHughes:
yeah I think it can work okay without that too. one mo

MeirionHughes:
Anything else needed?

guybedford:
Yes I wanted to test it. I tried it out and this works and bundles fine without the shim in jspm 0.17. I take it the reason you needed it is because you are running jspm 0.16 here?

MeirionHughes:
ahh perhaps. let me update my global jspm.

guybedford:
The upgrade path is breaking, so I certainly wouldn't force that on you. Happy to merge this as-is, let me know.

MeirionHughes:
yeah, if you don't mind. I think most people will be on 0.16 if they do `npm install jspm -g`

guybedford:
One other thing in this override - `batch-processor` is still a dependency, but doesn't need to be using the dist version here. This can be handled by amending this with `"dependencies": {}`.

MeirionHughes:
unfortunately, it would appear the shim doesn't work properly; I had been using it with the bundler (which worked). When used without any bundling (aurlia skeleton app). It complains: 

> uncaught exception: ReferenceError: utils is not defined
> 	Evaluating http://localhost:9000/jspm_packages/npm/element-resize-detector@1.1.0/dist/element-resize-detector.js

simonfox:
We are seeing the same issue since this was merged. Can we get this rolled back? We too were using this in Aurelia without issue prior to the override being added.

MeirionHughes:
Requested rollback. 
I'd be interested to know how you got it to work without the shim; it complains it can't find "utils" for me. :/

simonfox:
@MeirionHughes we didn't need to do anything special....just `jspm install npm:element-resize-detector` and then usage (abridged)...

```js
import erd from 'element-resize-detector';

export class ThatUsesResizeDetection {
  constructor(element, eventAggregator, singleEventQueue) {
    this.elementResizeDetector = erd({ strategy: 'scroll' });
  }
}
```

MeirionHughes:
hmm is that jspm 0.17?

simonfox:
@MeirionHughes nope, `0.16.30`

MeirionHughes:
how odd; all works fine on my home machine. my work machine was complaining about the "utils" during bundling. I'll have to update everything and try again. Sorry the grief. :P

simonfox:
@MeirionHughes no worries :+1:

MeirionHughes:
you'd be happy with the package alias though right? 
https://github.com/jspm/registry/pull/819/commits/f6f95957ccd017d5b146213d36506d86733cd38c

simonfox:
@MeirionHughes yes definitely

PR #834, opened 2016-03-23
override nodeca/pako main

MeirionHughes:


guybedford:
Did you try `jspm install npm:pako` here?

MeirionHughes:
ah no, hadn't realized it was on there; I can try when I get back to work.

PR #835, opened 2016-03-23
fengyuanchen/cropperjs override and mapping

tsop14:
@fengyuanchen


guybedford:
This looks great, sorry for the delay here.

npm is the preferred registry for packages when a package is available equally on npm and GitHub. Perhaps try switching this over to `npm:cropperjs`? If anything different happens you can add `jspmNodeConversion: false` to the configuration to ensure this behaves the same as the github version.

package-overrides/github/fengyuanchen/cropperjs@0.7.0.json
```diff
@@ -0,0 +1,16 @@
...
+  },
+  "shim": {
+    "dist/cropper": [
+      "./cropper.css!"
+    ]
+  },
+  "files": [
+    "dist/cropper.js",
+    "dist/cropper.css"
+  ]
```
guybedford: It can be advisable to leave this out for the npm package directory as it contains so few files. It can be nice to have the original scss folder for some users.

tsop14:
@guybedford Have swapped to npm. Let me know how it looks. Thanks.

guybedford:
Actually that `system-css` package is not official at all. Thanks, I've removed myself from that and added a deprecation message there. Correction added in https://github.com/jspm/registry/commit/cfd3be425634a93698c839ccd2289d89bcb7962a.

PR #837, opened 2016-03-24
Added SpoonX aurelia entries

RWOverdijk:
Added authentication, orm, api and notification. Makes installation a lot nicer for our users :)


guybedford:
:+1: thanks for your patience, I've been travelling... if there are registry delays like this feel free to ping others with registry access.

Issue #697, opened 2015-12-29
help for handsontable


sandys:
here's a small testcase I built - https://github.com/sandys/systems-handsontable

the only fix I see is to use handsontable via a direct <script> include - https://github.com/sandys/systems-handsontable/tree/fix (this was based on a suggestion here - https://github.com/handsontable/handsontable/issues/2906#issuecomment-167149256).

P.S. I asked a more generic question here - https://github.com/systemjs/systemjs/issues/989
Handsontable is an incredibly important component that is very hard to replicate. But it also has its own plugin system (https://github.com/handsontable/handsontable-ruleJS) that needs to work with it.  We have found it absolutely impossible to make it work with systemjs/jspm. The question is what to do in these cases - will systemjs be restricted to using the newer written libraries ?

guybedford:
I managed to get this to work by updating your package.json to the following:

```json
{
  "name": "hands-test",
  "version": "0.0.1",
  "main": "js/index.js",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1",
    "start": "jspm-server --port=4001 --no-browser"
  },
  "jspm": {
    "dependencies": {
      "babel-core": "npm:babel-core@^6.3.15",
      "babel-preset-es2015": "npm:babel-preset-es2015@^6.3.13",
      "babel-preset-react": "npm:babel-preset-react@^6.3.13",
      "babel-preset-stage-0": "npm:babel-preset-stage-0@^6.3.13",
      "babel-register": "npm:babel-register@^6.3.13",
      "crypto-js": "npm:crypto-js@^3.1.5",
      "formula": "npm:formula@^0.0.2-beta2",
      "handsontable/handsontable": "github:handsontable/handsontable@^0.20.2",
      "history": "npm:history@^1.13.1",
      "immutable": "npm:immutable@^3.7.6",
      "lodash": "npm:lodash@^3.10.1",
      "moment": "npm:moment@^2.2.0",
      "react": "npm:react@^0.13.3",
      "react-router": "npm:react-router@^1.0.2"
    },
    "devDependencies": {
      "babel": "npm:babel-core@^5.8.24",
      "babel-runtime": "npm:babel-runtime@^5.8.24",
      "core-js": "npm:core-js@^1.1.4"
    },
    "overrides": {
      "github:handsontable/handsontable@0.20.2": {
        "main": "dist/handsontable.full.js",
        "format": "global",
        "shim": {
          "dist/handsontable.full": {
            "deps": [
[...]

PR #838, opened 2016-03-24
Add Handsontable to the registry and package-overrides.

jansiegel:


package-overrides/github/handsontable/handsontable@0.24.1.json
```diff
@@ -0,0 +1,14 @@
+{
+  "main": "dist/handsontable.full.min",
```
guybedford: It's best to use the unminified dependency for best debugging support.

guybedford:
Is anything from the discussion in https://github.com/jspm/registry/issues/697 relevant here? That seemed to include dependencies on zeroclipboard, pikaday, moment and jquery. Are any of those needed by Handsontable for the default import?

jansiegel:
> Is anything from the discussion in #697 relevant here? That seemed to include dependencies on zeroclipboard, pikaday, moment and jquery. Are any of those needed by Handsontable for the default import?

I think it should work fine. Our main dependencies (pikaday, moment.js and ZeroClipboard) are all included in the `full` build - maybe except for the ZeroClipboard's `.swf` file, but I don't think there's a way around that. As for the jQuery, since about a year ago we're jQuery-free, so it shouldn't be and issue.

As for the RuleJS library, it's not really a Handsontable plugin - rather a separate, currently not fully supported repository, so I wouldn't treat it as an integral part of the Handsontable package.

package-overrides/github/handsontable/handsontable@0.24.1.json
```diff
@@ -0,0 +1,14 @@
+{
+  "main": "dist/handsontable.full",
+  "dependencies": {
+    "css": "0.1.0"
```
guybedford: `"css": "*"` is advisable here.

guybedford:
Ok that all looks great to go then pending the css comment.

Also once this is merged, can also be advisable to include this configuration under `"jspm"` in the Handsontable package.json itself so that can become the source of truth but no pressure here at all.

jansiegel:
> Also once this is merged, can also be advisable to include this configuration under "jspm" in the Handsontable package.json

Yup, it's probably a good idea, we'll take it into consideration with our next releases :)

guybedford:
:+1:

ineentho:
This does not work with jspm .12, there are several issues:
- The css file points to the wrong path (changing it to ./handsontable.full.css works, so no big deal)
- Something with traceur $defineProperty not being defined.

It works when importing with System.import('handsontable'), but not within a esm module. My guess is there is some conflict between babel and traceur, but I didn't figure out what was wrong.

Test case is a clean jspm project:

```
mkdir httest && cd httest && npm init -y && npm i -S jspm@beta && jspm init -y && jspm i handsontable
echo '
<script src="jspm_packages/system.src.js"></script>
<script src="jspm.browser.js"></script>
<script src="jspm.config.js"></script>
<script>
    System.import("httest")
</script>
' >> index.html
echo "import 'handsontable'" >> src/httest.js
```

I did some digging, but I know very little about how the handsontable bundle is made.

Issue #839, opened 2016-03-25
esprima package override


hmalphettes:
Hi everyone, we are currently looking at making `jspm install esprima` behave better: not install the nodejs schims for the `fs` and `process` modules.

It seems we might end-up adding a package override here.

Your expert input would be much appreciated if in fact the best solution is to change esprima's package.json:
https://github.com/jquery/esprima/issues/1486#issuecomment-200919022

Let me know if we are better off with a package override here.

Thanks for your attention!

ariya:
Note that as a library, Esprima does not depend on any modules (i.e. it is self contained).
However, included in its package is a command-line tool that relies on `fs` and `process`.

guybedford:
As posted, a `jspmNodeConversion: false` override will likely be what is wanted here in the registry.

hmalphettes:
OK, noted.

PR #845, opened 2016-03-28
Esprima does not need the `fs` and `process` shims

hmalphettes:
It is standalone and packaged as a UMD.

Fixes https://github.com/jspm/registry/issues/839
Reference: https://github.com/jquery/esprima/issues/1486

### DIFF
--- a/package-overrides/npm/esprima@2.7.2.json
+++ b/package-overrides/npm/esprima@2.7.2.json
@@ -1,0 +1,3 @@
+{
+  "jspmNodeConversion": false
+}


guybedford:
It would also be worth posting a PR like this to the jspm registry to get something in for the older versions. The check-in here is still preferred as it avoids having to store this config in the local package.json of every users' project that uses Esprima.

hmalphettes:
@guybedford I apologise for taking time to understand how to improve on this PR.
I read as much as I could the content of the jspm/registry but I am missing how to add support for older versions of esprima nicely.
Can you give me another hint or example to make this happen?
Thanks!

guybedford:
@hmalphettes my apologies I incorrectly read this as an Esprima project PR, so was suggesting you do a PR here. Let me know if it all works out ok.

hmalphettes:
No problem. Many thanks again for this and everything around jspm/systemjs.

PR #846, opened 2016-03-28
Angular override correction

guybedford:
Here's a correction to the previous override.

This:
* Updates versions for beta.10+
* Updates the dependency on zone.js and reflect-metadata in jspm 0.17 to be automatic for all files of angular (making angular-polyfills.js only necessary in jspm 0.16).
* Fixes up the zone.js override to correctly load Crypto when running in Node, without having to install all of crypto browserify.


amcdnl:
@guybedford angular is on beta12 at the moment. probably should bump the version here.

PR #856, opened 2016-04-06
Create angular-aria@1.5.0.json

subesokun:


package-overrides/npm/angular-aria@1.5.0.json
```diff
@@ -0,0 +1,15 @@
+{
+  "format": "global",
+  "jspmNodeConversion": false,
+  "main": "angular-aria",
+  "dependencies": {
+    "angular": "npm:angular@^1.5.0"
```
guybedford: The issue here is that the ecosystem is currently still stuck on loading Angular from github. It may be best to just use the GitHub version of Angular for all of these to avoid compatibility issues.

subesokun: As this PR won't influence the default github (i.e. bower) based ecosystem for Angular I thought that this change shouldn't cause compatibility issues. By introducing the Angular npm namespace we've just to make sure that each overwrite sticks explicitly to the npm ecosystem regarding Angular as otherwise you'll end up with a bower and npm installation of Angular in your application. But so far my published overwrites are respecting that.

If this changes are too critical for you I can add them also in my private jspm registry but I just thought that it might help others who want to use the npm version of Angular as well.

guybedford: Yes the point is that the whole ecosystem has to agree on a main source for angular, otherwise we end up with Angular being duplicated which should not happen.

subesokun: Ok makes sense. The main reason why I've switched to npm was that to avoid hitting github rate limits (though this should be no longer an issue) and that I can use our internal NPM registry for caching those external packages so that an installation works also "offline". If I rely on the Github repos I've somehow mirror these repos on demand which is quite complicated.

So for now I'll place these npm overwrites in my private jspm registry so that I can install all packages from the internal NPM registry.

guybedford: Sure that sounds sensible.

PR #859, opened 2016-04-06
Create moment@2.0.0.json

subesokun:
Avoid that a node conversion gets performed to reduce the number of npm dependencies.


guybedford:
The only risk here is if Moment makes changes in future that stop this override from still being the case. (eg directory requires, using Buffer etc). But we can consider a patch if that ever happens.

PR #864, opened 2016-04-07
Add knockout-sortable

ricardobrandao:


package-overrides/github/rniemeyer/knockout-sortable@0.12.0.json
```diff
@@ -0,0 +1,15 @@
+{
+    "directories": {
+        "lib": "build"
```
guybedford: We don't add `directories.lib` overrides to the registry directly as it is a breaking change. If you can update this without it that would be great.

ricardobrandao:
@guybedford done

PR #872, opened 2016-04-18
add github:npbenjohnson/plugin-svg

npbenjohnson:
Works like plugin-text via translate for embedding or loading svgs


guybedford:
Amazing work! This is great. Can we add this to the SystemJS plugins list as well at https://github.com/systemjs/systemjs#plugins?

PR #883, opened 2016-04-25
register potion-client

rolandjitsu:
@guybedford do you think we can merge this?


rolandjitsu:
@guybedford do you think we can merge this?

rolandjitsu:
I no longer need this patch.

PR #889, opened 2016-04-30
Update ui-router in registry.json 

christopherthielen:
Switch existing ui-router projects to NPM.
Add new ui-router projects


guybedford:
The only worry here is we're now splitting the Angular ecosystem through jspm - some packages using `npm:angular` and others using `github:angular/bower-angular`. I'd really prefer if we can stick with the GitHub repos, otherwise if we want to switch to npm for Angular 1, that needs to be an explicit decision for the whole ecosystem.

christopherthielen:
Hey Guy,

> The only worry here is we're now splitting the Angular ecosystem through jspm

Note: we made the choice to move towards npm and away from bower because ui-router is no longer an angular-1-only framework.

Currently, if our ng1 users `jspm install angular-ui-router@1.0.0-alpha.4`, they get a broken package.  The github tag for `1.0.0-alpha.4` is not packaged, it's only a tag of the source code we later use to create packages. 

The `angular-ui/ui-router` repo hosts framework-agnostic code, and framework-specific adaptors (currently for angular1 and angular2).  We are publishing [three npm packages](https://github.com/angular-ui/ui-router/tree/master/packages) from that one repo. 

- `npm:ui-router-core`: platform agnostic
- `npm:ui-router-ng2`: angular 2
- `npm:angular-ui-router`: angular 1

We created a separate repository to support our bower users https://github.com/angular-ui/angular-ui-router-bower but moving forward we will not be adding any additional bower support.

guybedford:
Thanks @christopherthielen for letting me know the plans here. It was an important decision we had to make to use GitHub for the Angular repos, but at the time the justification was exactly that since Angular 1 is supporting Bower, this would remain a good option for the jspm installs.

Is it not possible to use the Bower route here as well with that repo? It would make things easier if we can hold out a little longer. Or do you think it would be best to make a shift now entirely to npm for Angular 1?

So something like - 

* `jspm install angular-ui-router` -> `github:angular-ui/angular-ui-router-bower`
* `jspm install ui-router-ng2` -> `npm:ui-router-ng2`

`ui-router-core` doesn't matter so much, but is that using a peer dependency on Angular or assumed peer dependency or just global access?

christopherthielen:
I guess I don't quite grasp the concern about mixing npm and github. 

Since it seems to be important to you and thought out, I don't have a problem with using `"angular-ui-router": "github:angular-ui/angular-ui-router-bower"` for ng1, and `npm:` for the rest.  We plan to continue packaging and publishing bower releases to `angular-ui-router-bower`. 

I'll update the PR accordingly.

christopherthielen:
> ui-router-core doesn't matter so much, but is that using a peer dependency on Angular or assumed peer dependency or just global access?

ui-router-core doesn't have any dependencies on angular, nor angular2

christopherthielen:
( @guybedford On a side note since I have your attention, just wanted to say THANK YOU for all you've done for the javascript ecosystem! )

guybedford:
Looks great! The reason for using github is that it avoids the jspm Node conversion process by default which will parse all modules to check for browserify deps etc, and also because we can then ensure we don't depend on `npm:angular` or other unnecessary npm dependencies.

Thanks so much, it's very much an uphill process, so it's really great to hear positive feedback!

Just let me know if I can assist at all further with these configurations.

registry.json
```diff
@@ -87,8 +87,8 @@
   "angular-tree-control": "github:wix/angular-tree-control",
   "angular-ui-grid": "github:angular-ui/bower-ui-grid",
   "angular-ui-layout": "github:angular-ui/ui-layout",
-  "angular-ui-router": "github:angular-ui/ui-router",
-  "angular-ui-router-extras": "github:christopherthielen/ui-router-extras",
+  "angular-ui-router": "github:angular-ui/angular-ui-router-bower",
+  "angular-ui-router-extras": "npm:ui-router-extras",
```
christopherthielen: I just realized this is wrong.  The package on npm and bower is named `ui-router-extras`, not `angular-ui-router-extras`

PR #894, opened 2016-05-08
add @angular overrides

te-je:
For the angular 2 rc.

### DIFF
--- a/package-overrides/npm/@angular/common@2.0.0-rc.1.json
+++ b/package-overrides/npm/@angular/common@2.0.0-rc.1.json
@@ -1,0 +1,7 @@
+{
+  "meta": {
+    "src/*.js": {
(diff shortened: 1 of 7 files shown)


guybedford:
Thanks for your patience here. This looks good to merge to me for the Angular 2 override support. I'm wondering if it is worth referencing the ES module sources here with `main: 'esm/index.js'` overrides? That would give us Rollup support for builds.

//cc @robwormald @jeffwhelpley @amcdnl

te-je:
I was getting an error ``Uncaught Module 1 not present`` when using ``main: 'esm/index.js'`` with static builds.

guybedford:
Sure, thanks, I just wanted to check the reasoning.

guybedford:
(deps are not supported for ES modules which may explain that!)

PR #896, opened 2016-05-09
systemjs-reloader

npbenjohnson:
New attempt at a reloader, adds clientside change debouncing, plugin loaded resource change detection, dynamic dependency parent tracking, recovery on import failure, static build for non-jspm users, Instance specific attachment, and hook for defining a custom module storage provider for unload/reload.

### DIFF
--- a/registry.json
+++ b/registry.json
@@ -499,6 +499,7 @@
   "svg.js":  "github:wout/svg.js",
   "systemjs-hot-reloader": "github:capaj/systemjs-hot-reloader",
   "systemjs-hot-reloader-store": "github:peteruithoven/systemjs-hot-reloader-store",
+  "systemjs-reloader": "github:npbenjohnson/systemjs-reloader",
   "t7": "npm:t7",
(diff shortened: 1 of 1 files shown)


guybedford:
Nice work! Out of interest, what issue in systemjs-hot-reloader inspired this?

npbenjohnson:
I really just wanted to fix reload of plugin loaded content, and allow for marking parents of dynamically loaded resources to support hot reloading svgs. Since the actual core function isn't too complex it was easier to just write es6 classes with separate responsibilities than cover regressions against existing single file solution. It should also make forward migration easier because the SystemJs access is all encapsulated in one class.
I think it covers:
capaj/systemjs-hot-reloader#14
capaj/systemjs-hot-reloader#32 
capaj/systemjs-hot-reloader#36 
capaj/systemjs-hot-reloader#49 
capaj/systemjs-hot-reloader#50 
capaj/systemjs-hot-reloader#74

guybedford:
Thanks, plugin support is a huge plus definitely!

PR #899, opened 2016-05-16
Move css dependency down a folder

ben657:
To match structure:
- dist
  - ag-grid.js
  - styles
    - ag-grid.css


guybedford:
Thanks. Was this a change in a recent version of ag-grid? If so, then we should create a new override that is set to that version rather. If it was a bug in exactly this version though, then this would be fine to merge.

ben657:
@guybedford I think it's been a bug the entire time, seems it was like that in [this version](https://github.com/ceolter/ag-grid/tree/4.0.4) too. This shim would work with 3.3.0+

guybedford:
Sure, thanks for confirming!

PR #903, opened 2016-05-21
Remove directories.lib override from bootstrap@4.0.0-alpha.2

frederikschubert:
This is needed to get the sass files in the bootstrap distribution.


guybedford:
Unfortunately this would be a breaking change... I was hoping we could do this change when the alpha branch ends (either for the major, beta or rc, depending on what Bootstrap does first). If you want to rename this to be a `4.0.0.json` major override now so we don't miss that transition, that would be really helpful.

In the mean time, perhaps just stick with a local override.

frederikschubert:
I am currently traveling through Samoa but will change the PR as you suggested as soon as I can!

frederikschubert:
Closing in favor of [this](https://github.com/jspm/registry/pull/912).

PR #911, opened 2016-06-01
correct protobuf.js alias inline with npm

MeirionHughes:
node module for this package is "protobufjs", "protobuf" relates to different project. Also typings has this as "protobufjs".  

Sorry for the faff on this, it only just occurred to me that name conflict between jspm and npm package names would make targeting commonjs and system all the more harder.


MeirionHughes:
btw, is it preferable to use npm in the future, rather than github, or does it not matter?

guybedford:
npm is generally preferable as most packages publish there primarily but it doesn't matter too much.

PR #916, opened 2016-06-15
Fix directories and jquery in bootstrap 4

jakeNiemiec:
- Bootstrap 4 is incompatible with jquery 3+
- Set lib to . so that sass files are installed

Test using this repo: https://github.com/jakeNiemiec/jspm_bootstrap

### DIFF
--- a/package-overrides/github/twbs/bootstrap@4.0.0-alpha.2.json
+++ b/package-overrides/github/twbs/bootstrap@4.0.0-alpha.2.json
@@ -1,14 +1,14 @@
 {
-  "main": "js/bootstrap",
+  "main": "dist/js/bootstrap",
   "files": null,
   "ignore": [
     "dist/js/npm"
   ],
   "directories": {
-    "lib": "dist"
+    "lib": "."
   },
   "shim": {
-    "js/bootstrap": {
+    "dist/js/bootstrap": {
       "deps": [
         "jquery",
         "tether"
@@ -17,7 +17,7 @@
     }
   },
   "dependencies": {
-    "jquery": "*",
+    "jquery": "npm:jquery@^2.2.4",
     "tether": "github:HubSpot/tether@^1.1.1"
   }
 }


guybedford:
We can't change the existing github override because that would be a breaking change. Rather approach we can make to not force a breaking path is to make `npm:bootstrap` version work as the main repo for bootstrap in jspm and change to that.

What exactly is not working with npm:bootstrap? What is the error? Let's fix it.

jakeNiemiec:
@guybedford It is broken right now, `bootstrap@4.0.0-alpha.2` requires jquery **below** 3.0.0. 
Using this as an example https://github.com/jakeNiemiec/jspm_bootstrap .

We are simply importing bootstrap https://github.com/jakeNiemiec/jspm_bootstrap/blob/master/src/index.js#L2

### Desired outcome:
image

Everything loaded. Notice that jquery was properly included on the page, exported be bootstrap.

#### Lets try `jspm i npm:bootstrap@4.0.0-alpha.2`
image

Jquery no longer gets properly exported, it is loaded on the page **after** index.js executes (this is beyond the scope of my understanding).

#### Lets assume the user will try to install or include jquery to fix this according to the error `jspm i jquery`
image
image
😞  same error

#### At this point lets assume the user know that they need to override the format in the loaded jquery
image

image

This is due (in part) to `"github:twbs/bootstrap@4.0.0-alpha.2" > "dependencies" > "jquery": "*",`

jakeNiemiec:
SO question reporting this problem http://stackoverflow.com/q/37823898/6091685

guybedford:
@jakeNiemiec thanks I've updated the Bootstrap override so that it uses jquery 2! It seems the jquery update broke a lot of packages (potentially all the places we have `jquery: '*'` actually). We did that to support jquery 1 and 2 at the same time, but it seems that has backfired. I'm wondering if it is worth just going through and bulk updating all `jquery: '*'` overrides to `jquery: '2'`...

For Bootstrap, you're points are completely valid, but I've just made too many breaking override changes in the past and it's not fun breaking peoples' projects. So I'd really appreciated if we can focus on getting the npm override to work.

PR #918, opened 2016-06-18
fix(Semantic-UI@2.0.4.json): Maintain jQuery 2.x.x in Semantic UI

davismj:
Currently, "*" dependency pulls in jQuery 3.x.x, which breaks Semantic UI. This restores the original "^2.1.4" behavior.

closes https://github.com/jspm/jspm-cli/issues/1918


guybedford:
👍

PR #922, opened 2016-06-22
Create angular-ui-router-bower@0.3.1.json

ben657:
Since this is the package specified by jspm:angular-ui-router

### DIFF
--- a/package-overrides/github/angular-ui/angular-ui-router-bower@0.3.1.json
+++ b/package-overrides/github/angular-ui/angular-ui-router-bower@0.3.1.json
@@ -1,11 +1,11 @@
 {
   "main": "release/angular-ui-router.js",
-  "meta": {
+  "shim": {
     "release/angular-ui-router.js": {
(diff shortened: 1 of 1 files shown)


package-overrides/github/angular-ui/angular-ui-router-bower@0.3.1.json
```diff
@@ -0,0 +1,11 @@
+{
+  "main": "release/angular-ui-router.js",
+  "meta": {
+    "release/angular-ui-router.js": {
```
guybedford: For jspm 0.16 compatibility it is still best to use the

```
"shim": {
  "release/angular-ui/router": {
    "deps": ["angular"]
  }
}
```

approach, at least until jspm 0.17 is out.

Also `angular` should probably be a `dependencies: { "angular": "*" }` as well?

guybedford:
Thanks for posting, and apologies for the delay here, just let me know if the feedback makes sense.

PR #926, opened 2016-06-27
Add aws-sdk@2.4.0 override.

skorlir:
Use aws-sdk `dist` bundle as a global.

### DIFF
--- a/package-overrides/npm/aws-sdk@2.4.0.json
+++ b/package-overrides/npm/aws-sdk@2.4.0.json
@@ -1,0 +1,8 @@
+{
+  "main": "dist/aws-sdk",
+  "format": "global",
+  "exports": "AWS",
+  "map": {
+    "process": "@empty"
+  }
+}
--- a/package-overrides/npm/aws-sdk@2.4.0.json
+++ b/package-overrides/npm/aws-sdk@2.4.0.json
@@ -1,8 +1,10 @@
 {
   "main": "dist/aws-sdk",
   "format": "global",
-  "exports": "AWS",
-  "map": {
-    "process": "@empty"
+  "jspmNodeConversion": "false",
+  "shim": {
+    "dist/aws-sdk": {
+      "exports": "AWS"
+    }
   }
 }


package-overrides/npm/aws-sdk@2.4.0.json
```diff
@@ -0,0 +1,8 @@
+{
+  "main": "dist/aws-sdk",
+  "format": "global",
+  "exports": "AWS",
+  "map": {
+    "process": "@empty"
+  }
```
guybedford: If you add `"jspmNodeConversion": false` then the map config becomes unnecessary here.

package-overrides/npm/aws-sdk@2.4.0.json
```diff
@@ -0,0 +1,8 @@
+{
+  "main": "dist/aws-sdk",
+  "format": "global",
+  "exports": "AWS",
```
guybedford: This isn't a valid property - it should either be jspm 0.17 metadata:

```javascript
  "meta": {
    "dist/aws-sdk.js": { "exports": "AWS" }
  }
```

otherwise it's probably best to use jspm 0.16 compatible shim:

```json
  "shim": {
    "dist/aws-sdk": { "exports": "AWS" }
  }
```

automatic detection applies though when neither is present, which is also ok.

skorlir:
Updated per your suggestions, @guybedford !

guybedford:
Great, thanks for the PR.

skorlir:
The same overrides work for aws-sdk@2.4.2, which was just released. It looks like, usually, you only keep one version of overrides around - should I make a new PR for 2.4.2?

guybedford:
Overrides apply for all versions semver compatible with the override. So you only need to worry about this for version 3.

skorlir:
@guybedford 

This was working moments ago, but suddenly it has stopped working?

The aws-sdk version is the same. I'd hate to think this was only working due to some kind of fluke on my system, but all of a sudden I'm getting a brand-new error message:

```
err  (SystemJS) /media/Data/dev/work/encurate/cognito-proof-of-concept/jspm_packages/npm/aws-sdk@2.4.2/dist/aws-sdk.js:16304
  global[AWS] = AWS;
                  ^
  ReferenceError: AWS is not defined
      at /media/Data/dev/work/encurate/cognito-proof-of-concept/jspm_packages/npm/aws-sdk@2.4.2/dist/aws-sdk.js:16304:17
      at Object.exports.runInThisContext (vm.js:54:17)
  Evaluating /media/Data/dev/work/encurate/cognito-proof-of-concept/jspm_packages/npm/aws-sdk@2.4.2/dist/aws-sdk.js
  Error loading /media/Data/dev/work/encurate/cognito-proof-of-concept/src/app.js
```

skorlir:
I've figured out part of it. There's a bad (auto-generated?) "map" override in the default aws-sdk@2.4.0.json that breaks the browser build. My previous "map" overwrote that, and now setting "map" to "{}" fixes it. `jspm run app` still fails, however, with the message above, where it was working before...

guybedford:
@skorlir perhaps try `jspm install npm:aws-sdk@2.4.0 -o` to ensure the override is refreshed from the registry for an existing project, otherwise the local override will continue to be used regardless of what is in the registry.

PR #927, opened 2016-06-27
Update overrides to overwrite bad "map".

skorlir:
If you don't set "map" to {}, you get

```json
"map": { "browser": "lib/browser.js" }
```

Which hopelessly breaks everything, as "lib/browser.js" contains calls to `require` that the browser can't handle.

`jspm run app` still errors, but at least the browser build works again.


guybedford:
This shouldn't be necessary with `jspmNodeConversion: false` present. Perhaps check https://github.com/jspm/registry/pull/926#issuecomment-229311468?

PR #932, opened 2016-07-01
Added auth0-lock@10.0.0-rc.1

darkyen:
This adds overrides for 10.0.0-rc.1


guybedford:
Sorry for the delay here!

darkyen:
Much appreciated I understand your pressure didn't know what the policy is so created the issue, I'll be making PR's for rc2 today and 10.0.0 when it comes this week (: they are pretty much the same, is there a way to do this easier? cause they need the same overrides for react

PR #947, opened 2016-08-01
Alter aurelia-froala-editor registry entry to point to npm:aurelia-froala-editor

florinpopescu:


florinpopescu:
Thanks :smiley:

PR #950, opened 2016-08-04
Mocha 3 support

oligot:
Mocha 3 has recently been released:
https://github.com/mochajs/mocha/blob/master/CHANGELOG.md

As the registry uses semver to resolve override, this commit adds
a new file `mocha@3.0.0.json` which is basically just a copy of
`mocha@2.5.3.json`.

Note that I've only tested this is the browser and not in Node.js


guybedford:
Thanks!

PR #952, opened 2016-08-10
Bootstrap 3.3.7 supports jQuery 3

BBosman:
Bootstrap 3.3.7 jQuery requirements:  [Bootstrap's JavaScript requires jQuery version 1.9.1 or higher, but lower than version 4](https://github.com/twbs/bootstrap/blob/0b9c4a4007c44201dce9a6cc1a38407005c26c86/dist/js/bootstrap.js#L15).

### DIFF
--- a/package-overrides/github/twbs/bootstrap@3.3.7.json
+++ b/package-overrides/github/twbs/bootstrap@3.3.7.json
@@ -1,0 +1,12 @@
+{
+  "main": "dist/js/bootstrap.js",
+  "shim": {
+    "dist/js/bootstrap": {
+      "deps": ["jquery"],
+      "exports": "$"
+    }
+  },
+  "dependencies": {
+    "jquery": "*"
+  }
+}


package-overrides/github/twbs/bootstrap@3.3.7.json
```diff
@@ -0,0 +1,24 @@
+{
+  "main": "js/bootstrap",
+  "shim": {
+    "js/bootstrap": {
+      "deps": ["jquery"],
+      "exports": "$"
+    }
+  },
+  "dependencies": {
+    "jquery": ">=1.9.1 <4"
```
guybedford: This version format is not supported in jspm - we only support semver ranges. Perhaps try just `jquery: '*'` here?

BBosman:
I've updated the PR based on your suggestion.

I'm wondering why jspm wouldn't understand `>=1.9.1 <4` as given version format though, as it's a valid semver range according to [the docs](https://github.com/npm/node-semver)?

package-overrides/github/twbs/bootstrap@3.3.7.json
```diff
@@ -0,0 +1,24 @@
...
+  "shim": {
+    "js/bootstrap": {
+      "deps": ["jquery"],
+      "exports": "$"
+    }
+  },
+  "dependencies": {
+    "jquery": "*"
+  },
+  "files": ["dist", "fonts", "js", "css", "less", "grunt", "LICENSE"],
```
guybedford: I'd suggest just leaving this out actually unless it is really necessary?

package-overrides/github/twbs/bootstrap@3.3.7.json
```diff
@@ -0,0 +1,24 @@
...
+  "systemjs": {
```
guybedford: This is a jspm-0.17 specific property.

For jspm 0.16 compatibility we should still use:

```
"shim": {
  "dist/js/bootstrap": {
    "deps": ["jquery"],
    "exports": "$"
  }
}
```

guybedford:
Thanks for the update.

BBosman:
I made a literal copy of `bootstrap@3.3.6.json` in the same directory for this PR and only modified the `jquery` dependency.

I don't mind making the requested changes, but don't the other files need those changes as well?

guybedford:
Ah I missed that. The `files` we can get away with, but not having the `systemjs` property is critical for jspm 0.16 support.

If you add this, that should be fine for future versions now.

BBosman:
I've updated the PR.

I also dropped the `files`, as a quick glance at the bootstrap `packages.json` showed that it was out of sync anyway.

guybedford:
Excellent thanks so much for your efforts here.

PR #957, opened 2016-08-18
Use npm instead of GitHub for Intl.js

PeterOrneholm:
This package requires to be built and packages to work, otherwise the references from index.js is wrong, therefore the npm version should be used.


guybedford:
👍

PR #961, opened 2016-08-25
fix bootstrap 3.3.7 and add jquery 3.1.0 overrides

AshleyGrant:
Bootstrap 3.3.7 is broken when installed via JSPM currently. This is due to two problems. First jQuery 3.1.0 needs an override to tell JSPM that it should be loaded as an AMD module. Second, the Bootstrap 3.3.7 seems to need the "files" section or else it doesn't work. 

The jQuery override matches the jQuery 2.x override. The Bootstrap override is the same as the one I submitted for 3.3.6, except that now it is no longer locked to jQuery 2.x.


guybedford:
Thanks for this fix. It completely slipped my mind in the recent update that we need the files override because of the files property already present in the bootstrap repo.

I guess when updating overrides it is important to copy in the previous override creator for review - will aim to do this in future for all overrides.

AshleyGrant:
>I guess when updating overrides it is important to copy in the previous override creator for review - will aim to do this in future for all overrides.

:+1:

rpkilby:
is this broken? Looking in my packages, there is no `js/boostrap.js` file. I'm having to override this to `dist/js/bootstrap.js`

dougludlow:
Seeing the same thing as @rpkilby, the main file should be `dist/js/bootstrap.js`.

PR #964, opened 2016-08-29
 	add override for material-design-lite

Gheoan:
`material-design-lite`'s `main` field is not recognized from github endpoint.

### DIFF
--- a/package-overrides/github/google/material-design-lite@1.2.0.json
+++ b/package-overrides/github/google/material-design-lite@1.2.0.json
@@ -1,0 +1,3 @@
+{
+  "main": "material.js"
+}


guybedford:
The alternative here would be to add a "main" field in the override for the material-light on github.

It may be better to use the github version as that will typically install less unnecessary dependencies in jspm.

Perhaps have a look at that and let me know if you're sure you still want to do this?

//cc @truongminh

Gheoan:
Right, I noticed that installing from the npm endpoint took longer because two additional dependencies were installed. I will change this pull request to override the `main` field.

Gheoan:
PR is ready.

truongminh:
That is right!
thank @Gheoan

package-overrides/github/google/material-design-lite@1.2.0.json
```diff
@@ -0,0 +1,3 @@
+{
+  "main": "dist/material.min.js"
```
guybedford: It's usually advisable to use the non-minified versions for development.

Gheoan: Why? The minified file has `sourceMappingURL`.

Gheoan:
I corrected the path to the main file. guybedford see me comment [here ](https://github.com/jspm/registry/pull/964#discussion_r77101053)about minified file.

guybedford:
@Gheoan because in development source maps still don't provide as a good of a debugging experience as having the original sources due to quirks of the source maps format. See http://fitzgeraldnick.com/2015/06/19/source-maps-are-insufficient.html for more info.

Gheoan:
Did the changes suggested by @guybedford. Ready for merge.

guybedford:
👍

PR #969, opened 2016-09-03
Added libraries

sahiljain112:


registry.json
```diff
@@ -371,6 +373,7 @@
   "jsnlog": "npm:jsnlog",
   "jsol": "github:daepark/JSOL",
   "jsplumb": "github:sporritt/jsPlumb",
+  "jspm-loader-css":"npm:jspm-loader-css",
```
guybedford: I'm not sure this is still supported? Is it still working in jspm 0.16? I'm not sure it does in jspm 0.17?

sahiljain112: Yes. You're right. I switched to systemjs/plugin-css for loading css.

guybedford: Ok thanks, it may be best to remove this then.

aluanhaddad: Otherwise this looks great.

Issue #976, opened 2016-09-19
Cannot import from @angular/core/testing


frederikschubert:
Angular updated their publishing directory structure since RC 6 and it is not possible anymore to import from `@angular/core/testing` as JSPM tries to fetch `jspm_packages/npm/@angular/core@2.0.0/testing.js` where it should be fetching `jspm_packages/npm/@angular/core@2.0.0/bundles/core-testing.umd.js`.
The main entry already points to `bundles/core.umd.js` and the mapping for testing should be analogous.

I tried to import from `@angular/core/testing/index` but that doesn't work either because then the imports from the successfully fetched core module to other modules like `@angular/compiler/testing` will fail.

Related to https://github.com/jspm/jspm-cli/issues/2080

guybedford:
Do you want to just import `@angular/core/bundles/core-testing.umd.js` then?

frederikschubert:
Yes. I would like to map `@angular/{core,platform-browser,compiler,etc.}/testing` to `@angular/core/bundles/{core,platform-browser,compiler,etc.}-testing.umd.js`.

guybedford:
Are you on jspm 0.17 or 0.16?

frederikschubert:
0.17.0-beta.28

guybedford:
Then you can edit `jspm_packages/npm/@angular/core@x.y.z.json` and add the line:

```json
"map": {
  "./testing": "./bundles/core-testing.umd.js"
}
```

and the same for the others.

frederikschubert:
Ah that works! I tried adding a mapping but didn't find any documentation on the syntax.
I guess that those mappings should be added to the registry? I would like to make a PR!
Thank you for your help! Greatly appreciate what you are doing for this project.

guybedford:
A PR with the necessary configs for Angular is welcome.

vigie:
I am on JSPM 0.16 and now when I install `@angular/core` a `testing.js` file is written with the following:

```/* */ 
module.exports = require('@angular/core@2.4.7/bundles/core-testing.umd/index');````

This causes module loading failures at runtime.

How can I prevent the `index` from being appended in this file? If I remove it manually it works fine.

Thanks

GinaLazar:
Hello, 
    I'm having the exact same issue but I use jspm 0.16. Is there any way to map to the correct file from the bundles folder instead of testing.js?

aluanhaddad:
@GinaLazar you could install it with an override but this should not be necessary.
[...]

PR #977, opened 2016-09-20
fix(@angular/*): add mappings for testing umd bundles

frederikschubert:
As [suggested](https://github.com/jspm/registry/issues/976#issuecomment-248043926) by @guybedford.
Closes https://github.com/jspm/registry/issues/976

### DIFF
--- a/package-overrides/npm/@angular/core@2.0.0-rc.1.json
+++ b/package-overrides/npm/@angular/core@2.0.0-rc.1.json
@@ -1,0 +1,7 @@
+{
+  "meta": {
+    "src/*.js": {
+      "deps": ["reflect-metadata", "zone.js"]
+    }
+  }
+}
--- a/package-overrides/npm/@angular/core@2.0.0-rc.1.json
+++ b/package-overrides/npm/@angular/core@2.0.0-rc.1.json
@@ -1,7 +1,0 @@
-{
-  "meta": {
-    "src/*.js": {
-      "deps": ["reflect-metadata", "zone.js"]
-    }
-  }
-}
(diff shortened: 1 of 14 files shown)


package-overrides/npm/@angular/core@2.0.0-rc.1.json
```diff
@@ -1,7 +0,0 @@
-{
-  "meta": {
-    "src/*.js": {
-      "deps": ["reflect-metadata", "zone.js"]
-    }
-  }
-}
```
guybedford: This file can be left here, to support existing users. The new one without the prefix will still take preference.

package-overrides/npm/@angular/http@2.0.0-rc.1.json
```diff
@@ -1,7 +0,0 @@
-{
-  "meta": {
-    "src/*.js": {
-      "deps": ["reflect-metadata", "zone.js"]
-    }
-  }
-}
```
guybedford: And also for this file.

package-overrides/npm/@angular/core@2.0.0-rc.1.json
frederikschubert: Should I add the files back and then squash?

frederikschubert: Done.

vigie:
This fix doesn't work for me as I have `defaultJSExtensions: true` so System ends up trying to load `./bundles/platform-browser-dynamic-testing.umd.js/index.js`

Is there any better way to correct this other than adding all my own local overrides?

PR #981, opened 2016-09-29
Added override for npm:jquery-countdown

mramiro:


package-overrides/npm/jquery-countdown@2.2.0.json
```diff
@@ -0,0 +1,6 @@
+{
+  "dependencies": {
+    "jquery": ">=1.7"
```
guybedford: This should be `^1.7`.

mramiro: That's exactly the concern I had with this. If I use the caret constraint it would go from jQuery 1.7 to 1.12.4 (as of now), but according to their docs, the plugin works also with jQuery 2.x and 3.x (https://github.com/hilios/jQuery.countdown/tree/2.2.0) and I didn't want to left those versions out.

It seems the the jspm registry doesn't support "greater/less than" constraints, but the npm one does, since this override works as it is but doesn't if I set "registry": "jspm" (I'm running JSPM 0.16.45)

What's the recommended solution for this cases?

guybedford: The best we have for this is simply `jquery: '*'`. This is tracked in https://github.com/jspm/jspm-cli/issues/1253.

mramiro: I see... I'll update the PR accordingly.

guybedford:
👍

PR #986, opened 2016-10-12
feat(@angular/*): use different main entry in production mode

frederikschubert:
When setting production to true, the main entry is set to the index.js file.
The file (like the other files in the src directory) is an ES5 JavaScript file
with ES6 module format. Using this file instead of the UMD bundle allows
for tree-shaking when building the package in production mode using
rollup.

### DIFF
--- a/package-overrides/npm/@angular/compiler@2.0.0.json
+++ b/package-overrides/npm/@angular/compiler@2.0.0.json
@@ -1,6 +1,10 @@
 {
   "meta": {
+    "index.js": {
+      "format": "esm"
+    },
     "src/*.js": {
+      "format": "esm",
       "deps": [
         "reflect-metadata",
         "zone.js"
@@ -8,7 +12,10 @@
     }
   },
   "map": {
-    "@angular/compiler": ".",
+    "./bundles/compiler.umd.js": {
+      "production": "./index.js",
+      "~production": "./bundles/compiler.umd.js"
+    },
     "./testing": "./bundles/compiler-testing.umd.js"
   }
 }
--- a/package-overrides/npm/@angular/core@2.0.0.json
+++ b/package-overrides/npm/@angular/core@2.0.0.json
@@ -1,6 +1,10 @@
 {
   "meta": {
+    "index.js": {
+      "format": "esm"
+    },
     "src/*.js": {
+      "format": "esm",
       "deps": [
         "reflect-metadata",
         "zone.js"
@@ -8,7 +12,10 @@
     }
   },
   "map": {
-    "@angular/core": ".",
+    "./bundles/core.umd.js": {
+      "production": "./index.js",
+      "~production": "./bundles/core.umd.js"
+    },
     "./testing": "./bundles/core-testing.umd.js"
   }
 }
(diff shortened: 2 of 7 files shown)


package-overrides/npm/@angular/compiler@2.0.0.json
```diff
@@ -1,14 +1,21 @@
...
     "src/*.js": {
+      "format": "esm",
       "deps": [
         "reflect-metadata",
         "zone.js"
       ]
     }
   },
   "map": {
-    "@angular/compiler": ".",
```
guybedford: Can you explain why this is removed?

package-overrides/npm/@angular/core@2.0.0.json
```diff
@@ -1,14 +1,21 @@
...
     "src/*.js": {
+      "format": "esm",
       "deps": [
         "reflect-metadata",
         "zone.js"
       ]
     }
   },
   "map": {
-    "@angular/core": ".",
```
guybedford: ""

package-overrides/npm/@angular/http@2.0.0.json
```diff
@@ -1,14 +1,21 @@
...
     "src/*.js": {
+      "format": "esm",
       "deps": [
         "reflect-metadata",
         "zone.js"
       ]
     }
   },
   "map": {
-    "@angular/http": ".",
```
guybedford: ""

package-overrides/npm/@angular/platform-browser-dynamic@2.0.0.json
```diff
@@ -1,14 +1,21 @@
...
     "src/*.js": {
+      "format": "esm",
       "deps": [
         "reflect-metadata",
         "zone.js"
       ]
     }
   },
   "map": {
-    "@angular/platform-browser-dynamic": ".",
```
guybedford: ""

package-overrides/npm/@angular/platform-browser@2.0.0.json
```diff
@@ -1,14 +1,21 @@
...
     "src/*.js": {
+      "format": "esm",
       "deps": [
         "reflect-metadata",
         "zone.js"
       ]
     }
   },
   "map": {
-    "@angular/platform-browser": ".",
```
guybedford: ""

package-overrides/npm/@angular/router@3.0.0-rc.1.json
```diff
@@ -1,14 +1,21 @@
...
     "src/*.js": {
+      "format": "esm",
       "deps": [
         "reflect-metadata",
         "zone.js"
       ]
     }
   },
   "map": {
-    "@angular/router": ".",
```
guybedford: ""

package-overrides/npm/@angular/compiler@2.0.0.json
guybedford: Ok, if you're sure, you just included this before.

frederikschubert:
I will try to test it with a 0.16 project. The speedup during development comes from the fact that only one UMD has to be loaded and not ~800 files so the transpilation is not the bottleneck.
It is almost impossible to test angular applications without the bundle on IE which loads only 2 files in parallel...

guybedford:
Ok sure that is understandable.

PR #987, opened 2016-10-14
Use same override for svgo@0.7.1 as 0.6.1

joeldenning:
svgo released a new version that still has not fixed https://github.com/svg/svgo/issues/504. Since the dependencies are not statically declared, we need this override. See the [similar override for 0.6.1](https://github.com/jspm/registry/blob/master/package-overrides/npm/svgo%400.6.1.json). This pull request adds three more plugins that have been written that were not part of 0.6.1:

"../../plugins/removeXMLNS"
"../../plugins/removeElementsByAttr"
"../../plugins/addAttributesToSVGElement"

### DIFF
--- a/package-overrides/npm/svgo@0.7.1.json
+++ b/package-overrides/npm/svgo@0.7.1.json
@@ -1,0 +1,47 @@
+{
+  "meta": {
+    "lib/svgo/config.js": {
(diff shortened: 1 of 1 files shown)


guybedford:
:+1:

PR #1009, opened 2016-12-03
Migrates jquery-ui to npm package

### LINKED ISSUE #1008: Use npm:jquery-ui instead of github:components/jquery-ui
Motivation:
1. This is consistent with how [jquery|https://github.com/jspm/registry/blob/master/registry.json#L350] is handled.
2. The shim doesn't provide any additional support.
3. The shim lacks a license.
4. On new version release of jquery-ui there isn't the need to wait on the owners of the components project.

gsmith85:
jquery-ui has been updated in the regristry to point to npm:jquery-ui
instead of github:components/jqueryui.

Further, package-overrides that pointed at the github:components version
of both jquery-ui and jquery have been updated to rely on the npm version
of the package.

Fixes #1008


guybedford:
@aluanhaddad do have any link to further details here? This is a risky change because it will potentially break things for a lot of users, but it would be nice to consolidate on the npm version of jQuery definitely... we just need to be sure it will be a stable change first.

aluanhaddad:
@guybedford Unfortunately I don't have links to specific issues. 
Upon further reflection I would say it is probably better to leave packages that are not related to jquery-ui out of this PR and revisit them on a case by case basis.

gsmith85:
My take is this:
1. `github:components/jquery` is basically asking for the latest version of `jQuery` per the components shim.  I think in this case there is very very low risk to simply swapping in `npm:jquery`.
2. Specific versions of `jQuery` are the bigger risk and primarily older versions that may not have a `package.json`.  But taking a second pass only `knockout-sortable@0.12.0.json` falls into this category and uses `@^2.1.4` which will match `2.2.4` which [looks good to go](https://github.com/jquery/jquery/tree/2.2.4).

But your call.  I certainly don't want to hose a bunch of other dependencies, just trying to be a good citizen and leave things better than I found them!

package-overrides/github/components/jquery@2.0.3.json
```diff
@@ -1,5 +0,0 @@
-{
-  "main": "jquery",
-  "files": ["jquery.js"],
-  "format": "global"
-}
```
guybedford: Is there a reason for removing these overrides?

package-overrides/github/components/jquery@2.1.0.json
```diff
@@ -1,3 +0,0 @@
-{
-  "main": "jquery"
-}
```
guybedford: ""

package-overrides/github/components/jqueryui@1.11.3.json
```diff
@@ -1,7 +0,0 @@
-{
-    "main": "jquery-ui.js",
-    "registry": "jspm",
-    "dependencies": {
-        "jquery": "*"
-    }
-}
```
guybedford: ""

It is better to leave old overrides in place I think, and if they are conflicts, then to create a newer override for newer versions that takes preference.

package-overrides/github/components/jquery@2.0.3.json
aluanhaddad: I don't think we so. They are harmless as there is no need to maintain them. I would say there is no gain for breaking someone's code in this case, as unlikely as that may be.

gsmith85:
Okay so I'm going to revert all `components/jqueryui@<some-version>.json` and otherwise it look like we're good.  I'll kick those changes up in short order.

package-overrides/github/components/jquery@2.0.3.json
gsmith85: Cool, am going to leave all the changes I made to `components/` for posterity.  My primary motivation for removing them was to just eliminate clutter and avoid confusion.

Is it worth leaving a README in this directory indicating that the `npm` variants should be preferred?

package-overrides/github/components/jquery@2.1.0.json
gsmith85: 👍

package-overrides/github/components/jqueryui@1.11.3.json
gsmith85: 👍

aluanhaddad:
@gsmith85 thank you, very much appreciated!

@guybedford this looks good, but if you could a last look through this that would be great.

guybedford:
LGTM

PR #1017, opened 2016-12-23
npm:debug update override to support 2.5.1

frankwallis:
to resolve https://github.com/jspm/jspm-cli/issues/2204 and https://github.com/visionmedia/debug/issues/394


frankwallis:
Possibly, but it sounds to me like 0.16 only supports ```jspmNodeConversion: false``` so it should stay as is until 0.16 is deprecated

aluanhaddad:
Thank you @frankwallis!

PR #1001, opened 2016-11-15
feat(lordfriend/nya-bootstrap-select): add override for lordfriend/ny…

davinkevin:
…a-bootstrap-select to have correct css

PR #1020, opened 2017-01-03
feat(nya-bootstrap-select): add github version of override

davinkevin:
I push on PR #1001 an error, so the override was only available for the NPM version of the lib.

### DIFF
--- a/package-overrides/github/lordfriend/nya-bootstrap-select@2.1.9.json
+++ b/package-overrides/github/lordfriend/nya-bootstrap-select@2.1.9.json
@@ -1,0 +1,13 @@
+{
+  "main": "dist/js/nya-bs-select.js",
+  "dependencies": {
+    "css": "jspm:css@*"
+  },
+  "shim": {
+    "dist/js/nya-bs-select": {
+      "deps": [
+        "../css/nya-bs-select.css!"
+      ]
+    }
+  }
+}


guybedford:
It's usually best to try and pick between npm and github. Is it worth altering the npm version to get that to work? Eg with `jspmNodeConversion: false`?

davinkevin:
I will try, but I think with the node version, the install will grab all the `angular` module from npm... and our project use mainly `angular` dep from github... 

For the `jspmNodeConversion`, I am not contributor of this lib, so I can't add this to the package.json (if I understand what you want me to do with the `jspmNodeConversion` flag)

guybedford:
@davinkevin ah right, that makes sense. The alternative would be to pr the npm version to use angular from github, but this is probably best for now.

davinkevin:
Thanks for the merge.

BTW, the npm version is scoped, so the correct naming for the file should be `@lordfriend/nya-bootstrap-select@2.1.9`

If I try the override with the NPM version (which doesn't cause extra-angular download :D ), I have this error : 

```
Downloading npm:@lordfriend/nya-bootstrap-select@2.1.9

warn Error on processPackageConfig
     Package.json dependency css set to jspm:css@*, which is not a valid dependency format for npm.
     It's advisable to publish jspm-style packages to GitHub or another registry so conventions are clear.

warn Error processing package config for npm:@lordfriend/nya-bootstrap-select.
```

So the override for the NPM version seems to be broken.

PR #1027, opened 2017-01-18
Leaflet.vector-markers needs to have leaflet in dependencies, otherwise resolving it in jspm fails

zakjan:


package-overrides/npm/Leaflet.vector-markers@0.0.6.json
```diff
@@ -1,4 +1,6 @@
 {
-  "dependencies": {},
+  "dependencies": {
+    "leaflet": "^1.0.0"
```
aluanhaddad: Thanks for the PR. This looks great, but for maximum compatibility the minimum version should be `"^0.7.0"` to match the requirements specified here https://github.com/hiasinho/Leaflet.vector-markers/blob/master/package.json#L33

That is of course there is an incompatibility with this version that you are aware of.

zakjan: Oops, missed that. Leaflet 1.0.x is stable for some time, but maintainer of this package haven't updated its leaflet major version yet. It works fine with it. What is the preferred solution for such situation?

zakjan: @aluanhaddad Version fixed to match the original package.

PR #1028, opened 2017-01-25
Add jQuery-Autocomplete registry and override

smajl:


package-overrides/github/devbridge/jQuery-Autocomplete@1.3.0.json
```diff
@@ -0,0 +1,13 @@
...
+  "files": [
+    "dist/",
+    "typings/jquery-autocomplete/",
+    "readme.md"
+  ],
+  "dependencies": {
+    "jquery": "*"
+  },
+  "shim": {
+    "jquery.autocomplete": ["jquery"]
```
aluanhaddad: I have tested this with both the latest stable and beta releases of JSPM and the `"shim"` section is not necessary as this is a well behaved package that only binds to a global `jQuery` as a last resort. Simply having jquery as a dependency cause the right behavior in both 2.2.4 and 3.1.1 releases

package-overrides/github/devbridge/jQuery-Autocomplete@1.3.0.json
```diff
@@ -0,0 +1,13 @@
+{
+  "files": [
+    "dist/",
+    "typings/jquery-autocomplete/",
```
aluanhaddad: Awesome that you included the typings, that is much appreciated. However, there is an issue. The typings directory includes a file named `jquery.autocomplete-tests.ts` which writes directly into the global scope.

package-overrides/github/devbridge/jQuery-Autocomplete@1.3.0.json
smajl: Thanks for the test, I have removed the shim.

package-overrides/github/devbridge/jQuery-Autocomplete@1.3.0.json
smajl: Changed, thanks for finding this out.

smajl:
Requested changes made. :)

aluanhaddad:
Thanks @smajl!

PR #1029, opened 2017-01-31
new official less plugin

amenadiel:
new official less plugin, according to https://github.com/systemjs/plugin-less/issues/10


guybedford:
Thanks again for your work here!

PR #1031, opened 2017-02-10
added textures.js

BuildingJarl:
Added [textures.js ](https://riccardoscalco.github.io/textures/) to registry.json


registry.json
```diff
@@ -25,6 +25,7 @@
   "text": "github:systemjs/plugin-text",
   "ts": "github:frankwallis/plugin-typescript",
   "tsv": "github:wpf500/plugin-tsv",
+  "textures": "npm:textures",
```
guybedford: This section of the registry is for plugins. Can you move this down to the next section at its alphabetical location?

PR #1039, opened 2017-02-27
Fix the path

nicklasb:
The path is incorrect and makes the library not work, at least for me, this fixes that problem.

### DIFF
--- a/package-overrides/github/twbs/bootstrap@3.3.7.json
+++ b/package-overrides/github/twbs/bootstrap@3.3.7.json
@@ -1,7 +1,7 @@
 {
-    "main": "js/bootstrap.js",
+    "main": "dist/js/bootstrap.js",
     "shim": {
-        "js/bootstrap": {
+        "dist/js/bootstrap": {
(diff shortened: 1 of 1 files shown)


guybedford:
Is this on jspm 0.16 or jspm 0.17? If I remember correctly the issue here is that what fixes one breaks the other.

nicklasb:
I am at jspm@0.17.0-beta.40.
But why is only one of *all* those packages?

aluanhaddad:
I am trying to figure out what changed. Bootstrap 3.3.7 was released on July 25, 2016 but this had been working up until a few weeks ago I believe.

nicklasb:
@aluanhaddad Well, @guybedford reverted a similar change earlier, because it broke 0.16. 
As 0.17 is not released, that would seem  to be a good call. 
Has 0.17 stopped adding "dist" to the path?

https://github.com/jspm/registry/commit/918910c3567c09412a8dc3725e5c128c2d724c20

PR #1040, opened 2017-03-05
add override for tslib to use module entry point

frankwallis:
Currently trying to use ```importHelpers``` option throws an error due to importing named exports from commonjs, see https://github.com/frankwallis/plugin-typescript/issues/199

### DIFF
--- a/package-overrides/npm/tslib@1.0.0.json
+++ b/package-overrides/npm/tslib@1.0.0.json
@@ -1,0 +1,4 @@
+{
+  "main": "tslib.es6.js",
+  "format": "esm"
+}


aluanhaddad:
@frankwallis I was playing around with this last night and ran into the same issues. I used the same override locally that you have added here and it works great. Thanks for submitting this!

AdamWillden:
This change broke our builds today. We transpile as a build step rather than on the fly. As a result of this PR it was trying to load ESM rather than cjs and failing to do so. I had to re-override this change which was quite frustrating waste of time for us figuring out what went wrong.

timfish:
@AdamWillden how did you re-override this issue?

AdamWillden:
@timfish Not at computer at moment. I modified the entry added to package json. Removed the 'main' line and changed the 'format' line to 'cjs'

frankwallis:
I have created https://github.com/jspm/registry/pull/1041 to resolve this.

PR #1041, opened 2017-03-06
fix back compatibility with jspm 0.16

frankwallis:
see https://github.com/Microsoft/tslib/issues/26#issuecomment-284441324

There is an issue with pointing at the ES module file because not everyone is using a transpiler.

I had to add "esModule": true to make it work with SystemJS@0.20, it seems that commonjs module.exports is hidden behind the 'default' property otherwise, which didn't seem correct to me. Shouldn't it be possible to import module.exports directly without using 'default' or setting 'esModule'?

### DIFF
--- a/package-overrides/npm/tslib@1.0.0.json
+++ b/package-overrides/npm/tslib@1.0.0.json
@@ -1,4 +1,4 @@
 {
-  "main": "tslib.es6.js",
-  "format": "esm"
+  "main": "tslib.js",
+  "format": "cjs"
 }


aluanhaddad:
We could import it using
```ts
import * as tslib from 'tslib'; 
```
and from a TypeScript perspective that might be expected to work but but my understanding of SystemJS@0.20 and the node eps is that the semantic implications of this are invalid and it is not therfore not allowed. @guybedford any insight on this would very helpful.

aluanhaddad:
As this is now broken again for jspm@0.17, I am trying to find and test a solution that will work for both branches.

guybedford:
@aluanhaddad the alternative to `__esModule` set in the original source is to add a meta configuration:

```
"meta": {
  "tslib.js": {
    "esModule: true
  }
}
```

this should work through the jspm 0.17 metadata override mechanism, as it passes all meta options directly to SystemJS (even though this was a recent SystemJS addition).

Let me know if that can work for you.

aluanhaddad:
@guybedford Excellent. I was not aware of the `"esModule": true` setting. Thank you.

PR #1043, opened 2017-03-17
add esModule marker for compatibility with SystemJS 0.20

aluanhaddad:
This allows for correct loading of tslib under jspm 0.17/SystemJS 0.20 while preserving compatibility with jspm 0.16/SystemJS 0.19.
See related discussion in #1040, #1041, and https://github.com/Microsoft/tslib/issues/26.

I have tested this under jspm@0.17.41 with SystemJS@0.20.10 and under jspm@0.16.53 with SystemJS@0.19.46.

### DIFF
--- a/package-overrides/npm/tslib@1.0.0.json
+++ b/package-overrides/npm/tslib@1.0.0.json
@@ -1,4 +1,9 @@
 {
   "main": "tslib.js",
-  "format": "cjs"
+  "format": "cjs",
+  "meta": {
+    "tslib.js": {
+      "esModule": true
+    }
+  }
 }


guybedford:
Lgtm //cc @frankwallis

aluanhaddad:
@frankwallis any change you can take a look at this?

frankwallis:
LGTM - in fact I thought I had already done this in the last commit but it appears not, maybe I just added a local override.

aluanhaddad:
@frankwallis, @guybedford I've found this does not work correctly when running jspm bundle in 0.17
Specifically, it `esModule` metadata seems to be ignored. In browser transpilation works perfectly.

guybedford:
@aluanhaddad thanks for letting me know. It would help to know how it is being ignored in the bundling process. Does it fail bundling? Or does it not get included in the package config in the bundle itself?

aluanhaddad:
@guybedford Sorry for not being more specific.

The bundling process completes successfully and the bundle in turn contains all the full package config.

The error happens at runtime. The named exports of tslib are not available on the imported module when the bundle is executed. It only exposes the `default`.

guybedford:
@aluanhaddad thanks for the clarification, in that case it sounds like this might be resolved by https://github.com/systemjs/builder/issues/775, which would ensure the bundle includes the `esModule` flag.

PR #1047, opened 2017-05-08
Added v-accordion

JSlain:


JSlain:
Thanks! That was fast.

PR #1050, opened 2017-05-12
Make sure pbkdf2 is loaded correctly

nillis:


guybedford:
:+1: thanks for tracking this down.

PR #1051, opened 2017-06-13
add typescript@rc

aluanhaddad:
ESNext Dynamic Import!

### DIFF
--- a/package-overrides/npm/typescript@rc.json
+++ b/package-overrides/npm/typescript@rc.json
@@ -1,0 +1,14 @@
+{
+  "browser": {},
+  "map": {
+    "buffer": "@empty",
+    "child_process": "@empty",
+    "fs": "@empty",
+    "path": "@empty",
+    "process": "@empty",
+    "readline": "@empty"
+  },
+  "dependencies": {
+    "source-map-support": "*"
+  }
+}


guybedford:
Awesome, I haven't had a moment to look into the implementation - will be interesting to see if they are emitting the System.register support for it.

aluanhaddad:
@guybedford I got a chance to look into this a bit recently and the results seem promising. Compiling the following code
```ts
import('jquery').then(({default: $}) => {
  $('div').text('您好');
});
```
into the `System.register` module format using TypeScript results in the following JavaScript
```js
System.register([], function (exports_1, context_1) {
    var __moduleName = context_1 && context_1.id;
    return {
        setters: [],
        execute: function () {
            context_1.import('jquery').then(({ default: $ }) => {
                $('div').text('您好');
            });
        }
    };
});

```
Does this look correct?

If I understand the intent of the spec correctly, it is possible for the behavior of a dynamic import to be affected by arbitrary dynamic context associated with the importing module which is why the import is transpiled into `context_1.import` as opposed to `SystemJS.import`, but I could be way off on that.

There is also a new module target `esnext` which results in retention of dynamic imports along with other ESM constructs which seems ideal for scenarios that pipe the results through plugin-babel or take advantage of Rollup in the SystemJS Builder (when dynamic import support is added).

Unfortunately, there appears to be bug on the type checking side that fails to apply the `--allowSyntheticDefaultImports` logic to dynamically loaded modules and therefore does not acknowledge that `'jquery'` will have a `default` property. This doesn't affect emit.

aluanhaddad:
I just re-read [your comment in the TS repo](https://github.com/Microsoft/TypeScript/issues/14495#issuecomment-285294096) and this does indeed appear to be correct.

guybedford:
Thanks for checking this out, great to hear it's working ok. Yes exactly the context allows relative module names to work out relative to the runtime parent.

> Unfortunately, there appears to be bug on the type checking side that fails to apply the --allowSyntheticDefaultImports logic to dynamically loaded modules and therefore does not acknowledge that 'jquery' will have a default property. This doesn't affect emit.

Oh dear, well hopefully this can be sorted out soon. I still wish TypeScript would just get on board with https://github.com/Microsoft/TypeScript/issues/16093#issuecomment-304263496.

aluanhaddad:
I absolutely agree. The good news is that the TypeScript team does seem to be recommending the `import =` syntax more and more frequently.

Unfortunately, there is a seemingly endless plethora of tutorials and even official documentation for libraries such as [moment.js](https://momentjs.com/docs/#/use-it/typescript/) that give bad advice to TypeScript users who continue to write more and more non standard code that will turn into broken code.

With respect to the async import type checking bug, I filed https://github.com/Microsoft/TypeScript/issues/17444.

aluanhaddad:
By the way, dynamic import is now available to TypeScript users in plugin-typescript as of https://github.com/frankwallis/plugin-typescript/pull/209 released in 7.1.0

PR #1052, opened 2017-06-20
Add office-ui-fabric-js overrides

AquilaSands:


guybedford:
If this is your own package, you can include this override configuration in the published package.json file directly.

guybedford:
Ah I see it isn't :) Sure, merging!

PR #1054, opened 2017-07-22
various package overrides

selfagency:
added overrides for:

+  `github:InteractiveAdvertisingBureau/AdBlockDetection`
+  `npm:flickity`
+  `npm:jquery-migrate`

### DIFF
--- a/package-overrides/github/InteractiveAdvertisingBureau/AdBlockDetection@master.json
+++ b/package-overrides/github/InteractiveAdvertisingBureau/AdBlockDetection@master.json
@@ -1,0 +1,4 @@
+{
+  "format": "global",
+  "main": "adblockDetector.js"
+}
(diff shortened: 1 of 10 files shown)


selfagency:
Huh, maybe it was an issue in the previous JSPM version? I just removed the global declarations from my jQuery plugins and they seem to be working. This was not the case in 0.16.0 to my recollection. Anyhow, I've updated the PR accordingly.

selfagency:
Actually, the jQuery migrate override may also be unnecessary or not preferable. There are two versions in the package: one that provides console log output for developers and one that does not. I have the override set to the one that does not. I should probably spike that one too, yeah?

aluanhaddad:
Regarding the jQuery migrate plugin, I think what you currently have is correct.

PR #1056, opened 2017-08-15
fix for bootstrap@4.0.0-beta

AlbertWillemsen-Centric:
added overrides for bootstrap@4.0.0-beta.
added popper.js to registry.

### DIFF
--- a/package-overrides/github/twbs/bootstrap@4.0.0-beta.json
+++ b/package-overrides/github/twbs/bootstrap@4.0.0-beta.json
@@ -1,0 +1,24 @@
+{
+  "main": "js/bootstrap",
+  "files": null,
+  "ignore": [
+    "dist/js/npm"
+  ],
+  "directories": {
+    "lib": "dist"
+  },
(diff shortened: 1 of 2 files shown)


package-overrides/github/twbs/bootstrap@4.0.0-beta.json
```diff
@@ -0,0 +1,24 @@
+{
+  "main": "js/bootstrap",
+  "files": null,
+  "ignore": [
+    "dist/js/npm"
+  ],
+  "directories": {
+    "lib": "dist"
```
aluanhaddad: In previous overrides for bootstrap the SASS files were included, for those who work directly with the source, for example by overriding bootstrap mixins.

To preserve this use case it would be best to remove the `"directories"` property and instead qualify `"main"` with `dist/`.

aluanhaddad:
It looks like the next beta release will include `Popper.js` in the bundle itself which will make this easier.

PR #1057, opened 2017-08-19
Add tslib to registry.json

aluanhaddad:


aluanhaddad:
The jspm specific installation instructions for this package have also been updated to reflect the simplified installation https://github.com/Microsoft/tslib/pull/35

PR #1059, opened 2017-09-06
add override for npm:bootstrap@4.0.0-beta

bbak:
Replace Tether with [Popper.js](https://popper.js.org/), see [bootstrap#22444](https://github.com/twbs/bootstrap/pull/22444)


aluanhaddad:
It looks like the next beta release will include `Popper.js` in the bundle itself which will make this easier.

PR #1063, opened 2017-10-25
added registry entry and overrides for webgazer.js

ivanbacher:
WebGazer.js is an eye tracking library that uses common webcams to infer the eye-gaze locations -
 https://github.com/brownhci/WebGazer


registry.json
```diff
@@ -636,6 +636,7 @@
   "vue": "npm:vue",
   "webcomponents": "npm:webcomponents",
   "webfont": "github:components/webfontloader",
+  "WebGazer": "github:brownhci/WebGazer",
```
aluanhaddad: I think name should be all lower cased to match the [project's package.json](https://github.com/brownhci/WebGazer/blob/master/package.json#L2). That also means we want to change the override's filename as well.

aluanhaddad:
Thanks, @ivanbacher!

PR #1065, opened 2017-10-25
fix travis ci failures due to incompatibility between dependency and specified node version

### LINKED ISSUE #1064: registry CI failure caused by transitive dependency update
The registry transitively depends on the https://github.com/hapijs/hoek package. Use of this package causes Travis CI to fail as the Node.js version currently specified in `.travis.yml` is [0.10](https://github.com/jspm/registry/blob/master/.travis.yml#L3).

See https://github.com/hapijs/hoek/issues/193 for information on the specific error (essentially use of future reserved word in strict mode).

aluanhaddad:
This fixes the CI failures on Travis caused by the use of hoek, a transitive dependency, that requires node >= 4.x.
This updates the node version used by the CI environment to 4.8.5.

Resolves #1064

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-- '0.10'
+- '4.8.5'
 branches:
   only:
   - master


guybedford:
Great, thanks @aluanhaddad!

PR #1082, opened 2018-07-04
socket.io-client update location of main for >2.1.1

KoltesDigital:
It seems the overrides are not taken into account when the major version changes, right? Because there's already the same override for @1.7.0.


marleyjustin08:
`
Duplicate of #`

marleyjustin08:
https://github.com/jspm/registry/pull/1082#issuecomment-706713057

PR #1085, opened 2019-03-06
Fix files override for d3@4

gavinaiken:
Without this change, installing `d3@4` creates a directory `d3@4.13.0` containing only a single file, `package.json`, with no script file.


guybedford:
Thanks for the fix.

